PR #5, opened 2021-05-23
Follow tutorial to scaffold an idiomatic Express API

calebschoepp:
Starting to scaffold out the project. I figure the backend can live in `api/` and the frontend can live `www/`. At a high level, I think that each of these should be their own npm packages that are independent of each other. In the future as we want some more tooling around running the parts together we can use npm workspaces to get the two to work in tandem.

## Changes for backend
- Express
- Sample endpoint
- Some tests with Jest
- Prettier configuration
- Configuration from environment 

## TODO
- Find a better logging setup
- Scaffold out frontend with CRA


cyrusdiego (approved):
LGTM, I added issues for front-end scaffolding and the logging as well.

armiantos (commented):
LGTM :+1: a couple things we might want to add:
- prettier to consistently use singlequotes or double quotes, indentation level (i.e. enforce 2 space indentation)
- Add eslint

PR #12, opened 2021-05-24
Setup continuous integration for api

### LINKED ISSUE #13: Setup continuous integration for api

calebschoepp:
## Changes
- Setup continuous integration for `api`.
- Disallow lint warnings... Too strict? I kind of like the strictness to keep our codebase squeaky clean.
- Fix warnings and modify tests as needed

## TODO
- Setup continuous integration for `www`.
- Cache `npm i` in actions

Tickets have been added

Closes #13

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -12,7 +12,8 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "format": "prettier --config ../.prettierrc --ignore-path .prettierignore --write .",
-    "lint": "eslint src --ext .js,.ts"
+    "format:check": "prettier --config ../.prettierrc --ignore-path .prettierignore --check .",
+    "lint": "eslint src --ext .js,.ts --max-warnings 0"
   },
   "author": "",
   "license": "ISC",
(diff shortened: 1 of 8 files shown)


api/package.json
```diff
@@ -12,7 +12,8 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "format": "prettier --config ../.prettierrc --ignore-path .prettierignore --write .",
-    "lint": "eslint src --ext .js,.ts"
+    "format:check": "prettier --config ../.prettierrc --ignore-path .prettierignore --check .",
```
armiantos, line 15: Doesn't lint already do this (i.e. any format warnings are already done in `npm run lint`)?

armiantos (commented):
One comment regarding `format:check`. Also is it normal to do build > test > lint? I usually do lint > build > test. Regarding strictness, I also prefer stricter rules so no warnings allowed is :+1:  from me

calebschoepp:
> Also is it normal to do build > test > lint? I usually do lint > build > test.

Are you just suggesting that I should change the order of the CI jobs? That makes sense to me.

armiantos:
Might be a good idea, I do linting first under the assumption that build may take a while and might be best to save resources if we can find early errors from linting

armiantos (approved):
LGTM!

PR #16, opened 2021-05-24
Cache node_modules in CI

### LINKED ISSUE #15: Cache node_modules in continuous integration
Takes ~25s to install dependencies for `api` so it wouldn't hurt to cache it.

armiantos:
- Track `package-lock.json` so that `npm install`s are consistent between machines
- `node_modules` in API are cached with `package-lock.json` being the hash key
- Adapted from [How to cache node_modules in Github Actions with Yarn](https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh)

Cache hit
image

Cache miss (It will restore last cache but update the cache as a post process)
image

Depends on #12 
Closes #15

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,14 +29,15 @@ jobs:
             ${{ runner.os }}-
       - name: Install dependencies
         run: npm install
+      # TODO cache the following step if it gets too slow
+      - name: Install dependencies
+        run: npm install
+      - name: Check linting
+        run: npm run lint
       - name: Compile project
         run: npm run build
       - name: Run tests
         run: npm run test
-      - name: Check formatting
-        run: npm run format:check
-      - name: Check linting
-        run: npm run lint
 
   www:
     name: Integrate www
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,8 +7,8 @@ on:
     branches: [master]
 
 jobs:
-  build:
-    name: Build
+  api:
+    name: Integrate api
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -16,7 +16,13 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v1
-      - name: Install Dependencies
+      - name: Install dependencies
         run: npm install
-      - name: Compile TypeScript
+      - name: Compile project
         run: npm run build
+      - name: Run tests
+        run: npm run test
+      - name: Check formatting
+        run: npm run format:check
+      - name: Check linting
+        run: npm run lint
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,8 @@ jobs:
         working-directory: ./api
     steps:
       - name: Checkout code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
+      # TODO cache the following step if it gets too slow
       - name: Install dependencies
         run: npm install
       - name: Compile project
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,51 @@
...
+        working-directory: ./api
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-api-node-modules
+        with:
+          path: '**/node_modules'
```
calebschoepp, line 24: Will this cache the `node_modules/` under `www/` too? Don't we just want `./api/node_modules`?

api/package.json
```diff
@@ -12,6 +12,7 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "format": "prettier --config ../.prettierrc --ignore-path .prettierignore --write .",
+    "format:check": "prettier --config ../.prettierrc --ignore-path .prettierignore --check .",
```
calebschoepp, line 15: This can be removed

.github/workflows/main.yml
```diff
@@ -16,6 +16,19 @@ jobs:
...
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: npm install
       # TODO cache the following step if it gets too slow
```
calebschoepp, line 32: Remove TODO

.github/workflows/main.yml
armiantos: Because the working directory is in `./api/node_modules` this will only target the current `node_modules`

.github/workflows/main.yml
```diff
@@ -16,6 +16,19 @@ jobs:
...
+          cache-name: cache-api-node-modules
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: npm install
```
calebschoepp, line 31: Am I missing something or are we just installing dependencies twice now?

calebschoepp (commented):
Couple comments but real close. Nice speedup!

.github/workflows/main.yml
armiantos: Good catch!

api/package.json
armiantos: My bad I was trying out git rebase and confused myself so I think I'll stick to merging master locally now 😂

.github/workflows/main.yml
armiantos: Removed 👍

.github/workflows/main.yml
calebschoepp: I might be mistaken, but I think the working directory is only set to `api` for `run` jobs not `use` jobs. Not a huge deal either way.

api/package.json
calebschoepp: lol

calebschoepp (approved):
Other than that confusion on working directory LGTM

.github/workflows/main.yml
calebschoepp: See this https://stackoverflow.com/a/63122434/10142748

armiantos: Ah good catch, I was wondering why it didn' trigger a cache miss in some cases, let me try this out. Thanks for pointing out the resource!

PR #17, opened 2021-05-24
Front end scaffolding 

### LINKED ISSUE #8: Front end scaffolding

cyrusdiego:
- CRA with TS template 
- Installed redux and react-redux 
- modified prettierc 

# TODO
- CI for www (ticket already created)

Closes #8


www/README.md
```diff
@@ -1,3 +1,47 @@
-# CompE+ www
+# CompE+ Front-End
```
calebschoepp, line 1: Nit: I kind of prefer `Frontend` compared to `Front End` or `Front-End`

www/README.md
```diff
@@ -1,3 +1,47 @@
-# CompE+ www
```
calebschoepp, line 1: Maybe not for this PR but at a high level is this how we want to structure the README b/c I'd like to mirror it in the api. IMO it is good to have a guide to each script but I also think there should be general stuff about setting up dev environment, local testing, etc.

www/package.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "compe-plus-www",
+    "description": "Front end for CompE+",
```
calebschoepp, line 3: Nit: Again I prefer `Frontend`

www/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "version": "0.1.0",
```
calebschoepp, line 4: Maybe bump the api version back to `0.1.0` too. Let's save `1.0.0` for when we have an MVP.

This opens a bigger question of whether we want to be updating these versions every time we change things. Not really necessary I guess b/c we won't be publishing packages but is kind of idiomatic.

www/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "private": true,
```
calebschoepp, line 5: Add this to api?

www/package.json
```diff
@@ -0,0 +1,52 @@
...
+            "not dead",
+            "not op_mini all"
+        ],
+        "development": [
+            "last 1 chrome version",
+            "last 1 firefox version",
+            "last 1 safari version"
+        ]
+    },
+    "devDependencies": {
```
calebschoepp, line 44: Shouldn't `typescript` be a dev dependency?

calebschoepp (commented):
Nice work. Left a couple comments

www/package.json
cyrusdiego: I didn't know this but when using CRA it puts it in `dependencies`, according to [this](https://github.com/facebook/create-react-app/issues/6180) there isn't much distinction with dev dependencies or regular dependencies for React apps.

www/README.md
cyrusdiego: When you say mirror api, do you mean like the headers and general content will be similat to api? Then we'd have another markdown for setting up dev env, local testing, etc.?

www/package.json
cyrusdiego: wait, was it `1.0.0`? it says `0.1.0` for me right now. I can't tell if I'm tripping loool

cyrusdiego: oh you said **api version** lol ok changing that now

www/package.json
calebschoepp: Works for me

www/README.md
calebschoepp: I'm saying I would like the structure of the api and www readmes to be similar.

I think as an example they should have sections like setup, local development, testing, code style, scripts, etc.

calebschoepp (approved):
LGTM!

www/README.md
cyrusdiego: Makes sense. I'll put this in the notion as part of the technical design section so we can keep it in mind.

PR #32, opened 2021-05-30
Design 1st draft of DB schema

calebschoepp:
Here is a screenshot of what I'm thinking for DB schema
Screenshot from 2021-05-31 20-53-20

In the file diff you will see an outline for the REST API endpoints and query parameters.


cyrusdiego:
**documents**
What does the `is_review` field signify? And I'm assuming the `state` will be like "Pending", "In Review", "Completed", etc. 

Also, how would back and forth messages attached to the document work? So when a user uploads their resume, they can add a message and that entry in the `documents` table will be filled appropriately. Then, when the reviewer reviews it, would they have a new entry in the `documents` table or would they just overwrite the `message` field in that document row? I'm assuming the latter as the row would just retain the file URL and any annotations would be made to the document itself, not the database. 

**users**
I think it's missing `specialty`? Or are we making `program` like discipline + specialty (if the specialty is applicable)?

At the moment, I don't think we're using profile pictures in the UI, but I'm assuming it's easy to pull from the Auth stuff and it would be a "nice to have" addition in the UI / profile for the user.

calebschoepp:
> **documents**
> What does the `is_review` field signify? And I'm assuming the `state` will be like "Pending", "In Review", "Completed", etc.

`is_review` signifies whether the document is uploaded by the reviewer or the reviewee. This may or may not be a useful field to have -- I just kind of put it down when I was spit-balling.

Having `document_states` is only really relevant if we are allowing the users to work on uploading a resume/review for awhile. Like if that is the case then we would have states like `creating` and `ready`. Potentially if we can't think of more than those two it would just be a boolean field. That being said, if reviews/resumes must be uploaded in one sitting than this state is irrelevant.

Having `resume_review_states` is relevant b/c reviews have a lifecycle. `waiting_for_reviewer`, `in_review`, `archived` etc.

> Also, how would back and forth messages attached to the document work? So when a user uploads their resume, they can add a message and that entry in the `documents` table will be filled appropriately. Then, when the reviewer reviews it, would they have a new entry in the `documents` table or would they just overwrite the `message` field in that document row? I'm assuming the latter as the row would just retain the file URL and any annotations would be made to the document itself, not the database.

I'm treating both resume uploads by the reviewee and reviews by the reviewer as documents. No overwriting. All copies of the resume and reviews would be kept around and associated with the review. So for example:

1. Sally the junior uploads a resume for review with a message (a new document is created).
2. Bob the senior reviews the resume and also leaves an associated message (a new document is created and file is the marked up original pdf)
3. Sally uploads a revised pdf with a message (a new document is created)
4. Bob does another review (a new document is created)
5. etc.

I considered having `resumes` and `reviews` as different entities in the DB but decided against it b/c as of right now there is nothing fundamentally different about them. Open to changing my mind on this if you think there is a reason they significantly differ.

> **users**
> I think it's missing `speciality`? Or are we making `program` like discipline + speciality (if the speciality is applicable)?

My thoughts were that `program` represented discipline + speciality. I figured that it made sense to make it one b/c it makes it easier to handle edge cases like CS students or disciplines without specialities etc. Open to changing my mind on this.

> At the moment, I don't think we're using profile pictures in the UI, but I'm assuming it's easy to pull from the Auth stuff and it would be a "nice to have" addition in the UI / profile for the user.

Yep, I just looked up what user info you can pull from Google OAuth and added it to the users schema.

calebschoepp:
@cyrusdiego @armiantos Update: I've translated the diagram into schema in the up and down migration files. Still want your feedback.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+/* users table */
+CREATE TABLE IF NOT EXISTS users (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    email TEXT NOT NULL,
+    ccid TEXT NOT NULL,
+    program TEXT NOT NULL,
+    year SMALLINT NOT NULL,
+    given_name TEXT NOT NULL,
+    family_name TEXT NOT NULL,
+    full_name TEXT NOT NULL,
```
calebschoepp, line 23: Unclear to me if this is how we want to do names. Just what Google [gives us](https://developers.google.com/identity/sign-in/web/people)

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+    photo_url TEXT,
```
calebschoepp, line 24: Letting this be NULL b/c they might not have a photo and that is more semantically clear than the empty string.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+CREATE TABLE roles (
+    name text NOT NULL PRIMARY KEY
+);
+
+INSERT INTO
+    roles (name)
+VALUES
+    ('reviewer'),
+    ('interviewer'),
+    ('admin');
```
calebschoepp, line 44: No `'user'` role b/c everyone is a user.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+
+/* user_roles table */
+CREATE TABLE user_roles (
+    user_id UUID NOT NULL REFERENCES users,
+    role TEXT NOT NULL REFERENCES roles,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
calebschoepp, line 50: No `updated_at` field b/c these should only ever be created or deleted.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+);
+
+INSERT INTO
+    resume_review_states (name)
+VALUES
+    ('creating'),
+    ('seeking_reviewer'),
+    ('reviewing'),
+    ('finished'),
+    ('cancelled');
```
calebschoepp, line 66: These states need refinement based on how the application will actually work.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+/* document_states table */
+CREATE TABLE document_states (
+    name TEXT NOT NULL PRIMARY KEY
+);
+
+INSERT INTO
+    document_states (name)
+VALUES
+    ('creating'),
+    ('ready');
```
calebschoepp, line 92: It's unclear to me if we actually need this. Only useful if the user can spend time editing a document before "pushing" it to the review.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+
+/* documents table */
+CREATE TABLE documents (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    message TEXT NOT NULL,
+    file_url TEXT NOT NULL,
```
calebschoepp, line 98: Should this be allowing NULL? Can you just post a message?

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+
+CREATE TRIGGER update_time_documents
+    BEFORE UPDATE
+    ON documents
+    FOR EACH ROW EXECUTE PROCEDURE update_modified_column();
+
+/* time_slots table */
+CREATE TABLE time_slots (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    at TIMESTAMPTZ NOT NULL,
```
calebschoepp, line 115: Are the interview slots always 1 hour long? Do we need to track a duration? That seems too complicated and I don't think we should let interviewers choose how long their interviews will be.

api/db/migrations/001_initial_schema.up.sql
armiantos: My gut leaning towards not sending anything to the server until the `Request for review` button is pressed.

api/db/migrations/001_initial_schema.up.sql
armiantos: Given that the model represents a document I don't think it should be able to just send a message. Might also help clarify if `message` was renamed to `note`?

api/db/migrations/001_initial_schema.up.sql
armiantos: For v1, I'd say just record the starting time and assume it'll take at most 1 hour long.

api/db/migrations/001_initial_schema.up.sql
```diff
@@ -1,51 +1,137 @@
...
+    BEFORE UPDATE
+    ON time_slots
+    FOR EACH ROW EXECUTE PROCEDURE update_modified_column();
+
+/* interviews table */
+CREATE TABLE interviews (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    time_slot UUID NOT NULL REFERENCES time_slots,
+    interviewee UUID NOT NULL REFERENCES users,
+    interviewer UUID NOT NULL REFERENCES users
```
armiantos, line 131: Does `interviewer` correspond to `time_slots.user_id`?

armiantos (commented):
Haven't read through the README thoroughly but the overall schema makes sense to me.

api/db/migrations/001_initial_schema.up.sql
calebschoepp: - This implies that the reviewee needs to upload their resume in one sitting. This is fine but we might want to support editing.
- What about future reviews of the resume? Without this state the reviewer will have to review the resume in one sitting. With this state they would be able to make some edits and come back to the page before submitting.

I think we need to talk about this more in person.

api/db/migrations/001_initial_schema.up.sql
calebschoepp: Yep agreed. `note` is much better too!

api/db/migrations/001_initial_schema.up.sql
calebschoepp: That is correct. This gave me the idea to put a constraint on the table that those must match.

cyrusdiego (commented):
Schema looks good to me as well, just had some comments below. Added an item in the next meeting's agenda to discuss the `creating` state for resume's as well so we can discuss whether or not we want to support this. Also read through your comments from my previous post and I think everything you said makes sense.

api/db/migrations/001_initial_schema.up.sql
calebschoepp: Mostly useful from the perspective of supporting people with unique names. See this fantastic [blog post](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) for more details. Kind of an edge case thing.

api/db/migrations/001_initial_schema.up.sql
calebschoepp: Yep that is roughly the idea with `creating`. Might be something we remove when we talk about it more in the meeting.

`canceled` is useful in some other scenarios too. For example imagine the scenario in which a reviewer loses their reviewer role mid review. What happens to the review? I would suggest it becomes canceled. There are other cases too.

api/db/migrations/001_initial_schema.up.sql
calebschoepp: Good point about emulating the application process. I think having a `creating` state is mostly useful from the reviewers side so that we can support reviewing a resume over multiple "user sessions".

api/db/migrations/001_initial_schema.up.sql
calebschoepp: :+1:

calebschoepp:
@armiantos @cyrusdiego So beyond addressing your comments here, should I just merge this? We can just make the adjustments to the schema as necessary after thursday? Do I have the LGTM?

cyrusdiego (approved):
Yeah I think we can merge this then discuss changes in the meeting. Good work on this :D

armiantos (approved):
LGTM! PR was really easy to review with the diagram. 🍻

PR #33, opened 2021-06-01
Auth integration

### LINKED ISSUE #25: Integrate Auth with frontend

armiantos:
## Overview
- Adds Auth0 integration
- Things to set-up in Auth0:
  1. Create an Auth0 application (this will automatically create a Google Console application)
  2. Disable manual database store (so that all auth flows go through Google)
  3. Register `localhost:3000` as a valid login, logout redirect URI, and web origin
  4. Add custom rule to filter to just `ualberta.ca` domains

## Screenshots
| Signing in | Signed in |
| --- | --- |
| Screenshot_20210531_210426 | Screenshot_20210531_210454 |

## Future work
- Spike on Auth0 integration with back-end (seems like we might even be able to cover roles using Auth0)

## References
- `Auth0ProviderWithHistory` is taken from [The Complete Guide to React User Authentication with Auth0](https://auth0.com/blog/complete-guide-to-react-user-authentication/)
- [Auth0 user roles](https://auth0.com/docs/users/view-user-roles) to be used in server-side

Closes #25


www/src/index.tsx
```diff
@@ -1,12 +1,19 @@
...
+import Auth0ProviderWithHistory from './providers/Auth0ProviderWithHistory';
+
 import './index.css';
 import App from './App';
 import reportWebVitals from './reportWebVitals';
 
 ReactDOM.render(
     <React.StrictMode>
-        <App />
+        <Router>
```
armiantos, line 12: `Auth0Provider` uses a `Redirect` component that needs a `BrowserRouter` as its parent.

armiantos (commented):
Helpful tips for PR review

www/.env.sample
```diff
@@ -0,0 +1,2 @@
+REACT_APP_AUTH0_DOMAIN=
+REACT_APP_AUTH0_CLIENT_ID=
```
calebschoepp, line 2: It is secure to store `CLIENT_ID` in environment? This just gets baked into the app when shipped to the user right so technically it is public?

www/.env.sample
```diff
@@ -0,0 +1,2 @@
+REACT_APP_AUTH0_DOMAIN=
```
calebschoepp, line 1: Are we planning to support different auth environments for staging and prod and local?

www/README.md
```diff
@@ -2,6 +2,12 @@
 
 This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
 
+## Authentication set-up
+
+1. Copy `www/.env.sample` to `www/.env`
+1. Navigate to `compe-plus` application in ![Auth0](https://manage.auth0.com/dashboard)
```
calebschoepp, line 8: Who has the account on auth0?

www/package.json
```diff
@@ -4,6 +4,7 @@
     "version": "0.1.0",
     "private": true,
     "dependencies": {
+        "@auth0/auth0-react": "^1.5.0",
```
calebschoepp, line 7: Cool, is this like a special namespace for auth0?

calebschoepp: Specifically the `@auth0` part

www/src/index.tsx
```diff
@@ -1,12 +1,19 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { BrowserRouter as Router } from 'react-router-dom';
```
calebschoepp, line 3: IMO we shouldn't alias this. I like just `BrowserRouter`. I don't see the point of the alias when there is no conflict here.

www/.gitignore
```diff
@@ -21,3 +21,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.env
```
calebschoepp, line 25: Maybe expand this to catch the other kinds of env files we might end up with in a React project. See [here](https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used) for a list of them.

www/src/App.tsx
```diff
@@ -1,8 +1,24 @@
 import React from 'react';
+import { GetTokenSilentlyOptions, useAuth0 } from '@auth0/auth0-react';
+
+import AuthenticateButton from './components/auth/AuthenticateButton';
 import logo from './logo.svg';
 import './App.css';
 
+// TODO: Remove dummy requests
+type tokenAcquierer = (options?: GetTokenSilentlyOptions | undefined) => Promise<string>;
```
calebschoepp, line 9: Nit: `s/Acquierer/Acquirer/g`

calebschoepp, line 9: Isn't the `| undefined` implicit to `options?`?

calebschoepp, line 9: Nit: Type should start with capital

www/src/components/auth/AuthenticateButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+        return <LogoutButton />;
+    }
+
+    return <LoginButton />;
+};
+
+BaseButton.propTypes = {
+    text: PropTypes.string.isRequired,
+    onClick: PropTypes.func.isRequired,
+};
```
calebschoepp, line 37: Asking b/c I'm naive with TS and React. Are prop types still needed when you have the typing of TypeScript already?

www/src/components/auth/AuthenticateButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const { logout } = useAuth0();
+    return <BaseButton text="Logout" onClick={() => logout()} />;
+};
+
+const LoginButton = () => {
+    const { loginWithRedirect } = useAuth0();
+    return <BaseButton text="Login" onClick={() => loginWithRedirect()} />;
+};
+
+const AuthenticateButton: React.FC = () => {
```
calebschoepp, line 24: Nit: Maybe a better name for this is `AuthenticationButton` or `AuthButton`. Better yet, I think this should just export a function that runs the correct login/logout logic as needed. Then separate components can leverage the function as needed. This would separate out auth logic and how we render it. Anyways, we probably don't need to worry about this with this prototype if you don't want.

www/src/providers/Auth0ProviderWithHistory.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useHistory } from 'react-router-dom';
+import { AppState, Auth0Provider } from '@auth0/auth0-react';
+
+const Auth0ProviderWithHistory: React.FC = ({ children }) => {
+    const domain = process.env.REACT_APP_AUTH0_DOMAIN ?? '';
+    const clientId = process.env.REACT_APP_AUTH0_CLIENT_ID ?? '';
```
calebschoepp, line 8: In the future we should probably move this to a config module like we do in the `api`. Let us just do stuff like `config.auth0Domain`.

www/src/providers/Auth0ProviderWithHistory.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
calebschoepp, line 1: What is the point of this component. If I understand correctly it just makes sure that we can go back to the original page after logging in?

calebschoepp (commented):
Okay couple things. First off, great work! This was a lot cleaner and simpler than I expected. I left a lot of nit level comments b/c this PR already feels like it is in the right direction and doesn't need to be totally reworked or something. So yay :+1: 

Going to test it locally and read up some more on Auth0 before I give the LGTM. Also, I think that we should validate that Auth0 will meet our backend roles requirement before we commit to it.

Good work Armi!

www/.env.sample
armiantos: Did not think of that when I wrote this but yes we should so that we could test without affecting deployment :+1:

www/.env.sample
armiantos: I mean the client has to request a token from Auth0 anyway. I think it's safe to store it in the `.env` generated from Github secrets.

www/README.md
armiantos: The compe club account will. I still need to get in touch with Cyrus to get the credentials

www/package.json
armiantos: Looking at [auth0 - npm](https://www.npmjs.com/package/auth0) it might not be a "namespace" for their other libraries

www/src/index.tsx
armiantos: Sounds good :+1:

www/.gitignore
armiantos: Oh cool, I never knew about these. Will add

www/src/App.tsx
armiantos: That vim syntax :rofl:

armiantos: It should be, it was the type that came up when I hovered over `getAccessTokenSilently` but I'll see if I can remove it

armiantos: Good catch!

www/src/components/auth/AuthenticateButton.tsx
armiantos: This is a `react-eslint` issue. There's an online debate whether this is needed. I personally would like to disable that rule if possible :rofl:

www/src/components/auth/AuthenticateButton.tsx
armiantos: The logic is all in `useAuth0` so yes I think once we get to actually integrating it with the UI, we'll make a separate PR to clean this up

www/src/providers/Auth0ProviderWithHistory.tsx
armiantos: Sounds good, let me merge that in here actually

www/src/providers/Auth0ProviderWithHistory.tsx
armiantos: I was imagining the case when the refresh token expires and an explicit re-sign in is required if the user has left the site on for a while. But then in that case we could also just redirect the user back to a fixed page (i.e. the dashboard)

www/.env.sample
calebschoepp: Sounds good

www/README.md
calebschoepp: Okay

www/src/components/auth/AuthenticateButton.tsx
calebschoepp: Okay well FWIW if we have the props explicitly typed somewhere with TS then I'm fine to remove it.

www/src/providers/Auth0ProviderWithHistory.tsx
calebschoepp: :+1:

www/src/components/auth/AuthButton.tsx
```diff
@@ -0,0 +1,39 @@
+import { useAuth0 } from '@auth0/auth0-react';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+type BaseButtonProps = {
```
cyrusdiego, line 5: Do we want to give unique names to Prop types or just name them `Props`? I think this is my coding style and I'm ok with either. Since they should be local to the file anyways they don't necessarily need a unique name aside from being verbose.

www/src/components/auth/AuthButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const { isAuthenticated } = useAuth0();
+
+    if (isAuthenticated) {
+        return <LogoutButton />;
+    }
+
+    return <LoginButton />;
+};
+
+BaseButton.propTypes = {
```
cyrusdiego, line 34: Is this required? I ask because I've never seen this tbh.

www/src/providers/Auth0ProviderWithHistory.tsx
```diff
@@ -0,0 +1,29 @@
...
+    };
+
+    return (
+        <Auth0Provider domain={domain} clientId={clientId} redirectUri={window.location.origin} onRedirectCallback={onRedirectCallback}>
+            {children}
+        </Auth0Provider>
+    );
+};
+
+Auth0ProviderWithHistory.propTypes = {
```
cyrusdiego, line 25: Same comment as above

www/src/components/auth/AuthButton.tsx
armiantos: I was able to get rid of this in this file by not using `React.FC` but this restricts us from making components that can render a `children` prop. Ideally I'd just disable this setting as mentioned in [React eslint error missing in props validation](https://stackoverflow.com/questions/38684925/react-eslint-error-missing-in-props-validation) what do you think?
image

www/src/components/auth/AuthButton.tsx
armiantos: I'm personally for keeping them unique just so I can see the type name in VSCode. I.e. 
image

cyrusdiego: Fair fair, I'll follow suit with this style in my PR

www/src/components/auth/AuthButton.tsx
armiantos: Was able to fix it, I thought typescript could infer the types from `FC` directly but seems like it needs us to explicitly type the props in the function itself. Thanks for showing me this 🍻

www/src/components/auth/AuthButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+type BaseButtonProps = {
+    text: string;
+    onClick: () => void;
+};
+
+const BaseButton: FC<BaseButtonProps> = (props: PropsWithChildren<BaseButtonProps>) => {
+    return <button onClick={props.onClick}>{props.text}</button>;
+};
+
+const LogoutButton = () => {
```
cyrusdiego, line 13: I just noticed, for consistency should this be typed (or does typescript infer this?)
```suggestion
const LogoutButton: FC = () => {
```

www/src/components/auth/AuthButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const { logout } = useAuth0();
+    return <BaseButton text="Logout" onClick={() => logout()} />;
+};
+
+const LoginButton = () => {
```
cyrusdiego, line 18: same as above 
```suggestion
const LoginButton: FC = () => {
```

cyrusdiego (approved):
Had one comment but I think this LGTM! Will most likely need to ask you stuff when I integrate my PR but should be fun :')

www/src/components/auth/AuthButton.tsx
armiantos: No its not inferred as a functional component, just a function that returns a JSX Element. I'll add this for consistency, thanks for catching this!

armiantos: Typed all components, let me know if I missed any 👍

PR #36, opened 2021-06-03
Setup Zapatos

calebschoepp:
## Changes
- Configure Zapatos
- Check in types for Zapatos
- Setup a pooling module to facilitate access to Postgres
- Improve logging
    - `VERBOSE` now will actually show up to `DEBUG` level logs
    - New mode `EXTRA_VERBOSE` will show up to `TRACE` level logs for the most granular information i.e. every step happening with Postgres
    - Log levels actually showing as strings in the logs now


api/src/controllers/sample.ts
```diff
@@ -1,15 +1,21 @@
...
  */
-const sampleHealthCheck = (req: Request, res: Response): Response => {
-    req.log.info('Sample health check route called.');
+const sampleHealthCheck = async (req: Request, res: Response): Promise<Response> => {
+    req.log.debug('Sample health check route called.');
 
-    return res.status(200).json({ message: 'pong' });
+    // TODO remove - this is just an example of how Zapatos can be used.
+    // In the future all DB access will be performed in the repository layer
+    const allUsers = await db.select('users', db.all).run(pool);
```
cyrusdiego, line 16: To confirm, the first parameter is the table name and the second parameter is a `WHERE` condition? And in this case, `db.all` just specifies no condition and gets all rows from the table?

api/package.json
```diff
@@ -14,7 +14,7 @@
         "test:watch": "jest --watch",
         "format": "prettier --config ../.prettierrc --ignore-path .prettierignore --write .",
         "lint": "eslint src --ext .js,.ts --max-warnings 0",
-        "psql": "psql -h localhost -U postgres -d compe-plus"
+        "psql": "PGPASSWORD=postgres psql -h localhost -U postgres -d compe-plus"
```
cyrusdiego, line 17: What does `PGPASSWORD=postgres` doe?

cyrusdiego (approved):
Just had some questions but I think this LGTM!

api/package.json
armiantos: I think it provides the password for the user postgres in `-U postgres` (so we don't have to type out the password everytime)

api/src/util/pool.ts
```diff
@@ -0,0 +1,18 @@
+import config from './config';
+import logger from './logger';
+import pg from 'pg';
+
+/**
+ * Pool of Postgres connections to avoid overhead of connecting on every request.
+ */
+const pool: pg.Pool = new pg.Pool({
```
armiantos, line 8: Not really a comment, just wanted to know what the convention will be forward. Do we want to explicitly type the variables even in obvious cases?

armiantos (commented):
Overall looks good to me. This is really cool! I've never heard of this library before. 

Curious where we get the database url in the zapatosconfig file though (or where they're being used)

api/src/controllers/sample.ts
calebschoepp: That is correct. `db.select` is a helper method to avoid re-writing some of the more common SQL queries. Alternatively we can use a format where we just pass in a full SQL query. I highly recommend reading the docs if you want to learn more about Zapatos

api/package.json
calebschoepp: Yep, it is setting an environment variable that psql uses to login. See [this](https://unix.stackexchange.com/questions/56444/how-do-i-set-an-environment-variable-on-the-command-line-and-have-it-appear-in-c)

api/src/util/pool.ts
calebschoepp: Umm idk. Probably not? I just did it here b/c it is an exported type - not really necessary though.

calebschoepp:
> Curious where we get the database url in the zapatosconfig file though (or where they're being used)

It is being used to access the DB when Zapatos scrapes it for schema.

PR #52, opened 2021-06-05
Update Postgres schema

### LINKED ISSUE #43: Remove document state

calebschoepp:
# Overview

Update Postgres schema to reflect conversations we had in June 3 meeting.

# Changes

- Removed concept of `document_state`
- Changed `resume_review_states` table into a `resume_review_state` enum and `roles` table into `role_type` enum. I did this because:
    - Users don't need to add new types so no need make it INSERTABLE like a table
    - Enum is simpler than inner joins
    - Gives us nicer typing with Zapatos i.e. we will have strict typing in the service on the names of the states and roles
- Add `IF NOT EXISTS` in a number of places to make it more idempotent
- Update Zapatos schema

# Questions & Notes

- This is not standard practice to modify an existing migration file. I really should create a new one - something like `002_remove_document_state`. I'm not doing this b/c sticking to one migration file right now simplifies things and we have no production DB to worry about yet.
- ~I want #36 to merge before this~ merged

# Issue tracking

Closes #43 

# Testing & Demo

- Verified that up and down migrations successfully runned
- Inspected what Zapatos schema would look like with enum


api/src/zapatos/.eslintrc.json
```diff
@@ -2,4 +2,4 @@
   "ignorePatterns": [
     "*"
   ]
-}
+}
```
calebschoepp, line 5: Want to keep this `No newline` here b/c it is auto-generated and we'll keep running into this issue.

api/db/migrations/001_initial_schema.down.sql
```diff
@@ -10,17 +10,15 @@ DROP TRIGGER IF EXISTS update_time_documents ON documents;
...
 
 DROP TABLE IF EXISTS resume_reviews;
 
-DROP TABLE IF EXISTS resume_review_states;
+DROP TYPE IF EXISTS resume_review_state;
 
 DROP TABLE IF EXISTS user_roles;
 
-DROP TABLE IF EXISTS roles;
+DROP TYPE IF EXISTS role_type;
```
armiantos, line 21: Any reason we're using `role_type` instead of just `role` since it's already a `TYPE`?

armiantos (commented):
Overall LGTM, let me know what you think of my one comment lol

api/db/migrations/001_initial_schema.down.sql
calebschoepp: I didn't want it to conflict with field names like `user_roles.role`. Open to a better name

armiantos: Good point, let's keep this 👍

armiantos (approved):
LGTM! Learning more about back-end everyday

PR #55, opened 2021-06-06
Set-up Auth0 secure route on back-end

### LINKED ISSUE #26: Integrate Auth with backend
- update db 
- register app as authenticated using Google console

armiantos:
# Overview

* Set-up secure routes using Auth0
* `api/secure/v1` routes require an API token requested by the client with an audience (or API identifier) and scope defined by us in Auth0 API section
* Give sample secure API call

# Changes

* Install `cors` server side (I'm not sure if this is safe to keep)
* Create jwt checker middleware to ensure client authenticity on secure routes
* Install `axios` on React client to make server requests
* Set-up environment variables for Auth0 variables

# Questions & Notes

Are there any open questions? Anything else important we should know?

# Issue tracking

Closes #26 

# Testing & Demo

Try `curl`ing the secure ping endpoint without a token

image

Pressing the button in the React app after signed in should give you the pong response in the browser console

image

# TODO

* Write tests for middleware


api/src/routes/middleware.ts
```diff
@@ -26,7 +28,22 @@ function notFound(): Middleware {
     };
 }
 
+/**
+ * Returns middle ware that will ensure the client accessing is authenticated.
+ * @returns JwtCheck middleware.
+ */
+function checkJwt(): Middleware {
```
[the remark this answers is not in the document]
calebschoepp: Another thing to note is that I would like to follow the pattern I outline in #56 with the `authorize` and `authenticate` endpoints. That is, if they fail they throw the appropriate `AuthenticationException` or `AuthorizationException` which is caught by the error handling middleware which will return the appropriate response.

Don't worry about throwing the exceptions in this PR though. #56 needs to merge first.

www/src/util/config.ts
```diff
@@ -1,13 +1,22 @@
 const AUTH0_DOMAIN = process.env.REACT_APP_AUTH0_DOMAIN || '';
 const AUTH0_CLIENT_ID = process.env.REACT_APP_AUTH0_CLIENT_ID || '';
 
+const SERVER_ENDPOINT = process.env.REACT_APP_SERVER_ENDPOINT || '';
+const SERVER_AUDIENCE = process.env.REACT_APP_SERVER_AUDIENCE || '';
```
calebschoepp, line 5: IMO these should have sane defaults for local development e.g. `const SERVER_ENDPOINT = process.env.REACT_APP_SERVER_ENDPOINT || 'http://localhost:1337';`

www/src/providers/Auth0ProviderWithHistory.tsx
```diff
@@ -17,7 +17,7 @@ const Auth0ProviderWithHistory: FC<Auth0ProviderWithHistoryProps> = ({ children
     };
 
     return (
-        <Auth0Provider domain={domain} clientId={clientId} redirectUri={window.location.origin} onRedirectCallback={onRedirectCallback}>
+        <Auth0Provider domain={domain} clientId={clientId} redirectUri={window.location.origin} onRedirectCallback={onRedirectCallback} audience={config.server.audience} scope={'call:ping'}>
```
calebschoepp, line 20: Fine for now, but those scopes should really become a TypeScript Enum.

Also question, eventually will the scopes this component needs to claim basically just be all possible scopes?

www/src/routes/ResumeReview.tsx
```diff
@@ -1,8 +1,31 @@
...
 const ResumeReview: FC = () => {
-    const { isAuthenticated } = useAuth0();
+    const { isAuthenticated, getAccessTokenSilently } = useAuth0();
+
+    useEffect(() => {
+        (async () => {
+            try {
+                const token = await getAccessTokenSilently({
+                    audience: config.server.audience,
+                    scope: 'call:ping',
```
calebschoepp, line 14: What happens if the user doesn't have the `call:ping` scope? Will it just get an access token but get `403`'s at the server. OR will it fail to get access token from Auth0?

www/src/routes/ResumeReview.tsx
```diff
@@ -1,8 +1,31 @@
...
+import axios from 'axios';
 import { useAuth0 } from '@auth0/auth0-react';
-import React, { FC } from 'react';
+import config from '../util/config';
 
 const ResumeReview: FC = () => {
-    const { isAuthenticated } = useAuth0();
+    const { isAuthenticated, getAccessTokenSilently } = useAuth0();
+
+    useEffect(() => {
```
calebschoepp, line 9: Maybe stick a `// TODO: Remove this` above this so we don't forget to take a random call to our api out of the code when we ship this.

calebschoepp (changes requested):
Amazing. Auth is looking easier and easier. Thanks for this work. Left a bunch of comments I'd like resolved before the approval but we are real close.

api/package.json
```diff
@@ -38,9 +39,13 @@
         "typescript": "^4.2.4"
     },
     "dependencies": {
+        "cors": "^2.8.5",
```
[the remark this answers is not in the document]
armiantos: Opened #65

api/package.json
```diff
@@ -38,9 +39,13 @@
...
         "dotenv": "^10.0.0",
         "express": "^4.17.1",
+        "express-jwt": "^6.0.0",
+        "express-jwt-authz": "^2.4.1",
         "express-pino-logger": "^6.0.0",
+        "jwks-rsa": "^2.0.3",
```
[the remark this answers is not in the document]
armiantos: There's no need to add `@types` on these, 2/3 of them already support native Typescript, namely

image
image

Do we want to explicitly add the type dependency for `express-jwt`? `jwks-rsa` already adds this dependency to us

api/src/routes/middleware.ts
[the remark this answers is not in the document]
armiantos: Sounds like a plan! Opened #66

api/src/routes/middleware.ts
```diff
@@ -26,7 +28,22 @@ function notFound(): Middleware {
...
+    return (req: Request, res: Response, next?: NextFunction): void => {
+        if (next === undefined) {
+            return;
+        }
```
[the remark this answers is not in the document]
armiantos: I did it because `jwtCheck` needs to have a next function

api/src/server.ts
```diff
@@ -15,6 +21,7 @@ router.use(express.json());
 
 /** Routes */
 router.use('/api/v1', sampleRoutes);
+router.use('/api/secure/v1', secureRoutes);
```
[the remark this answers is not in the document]
armiantos: Agreed

api/src/util/jwtCheck.ts
```diff
@@ -0,0 +1,16 @@
+import jwt from 'express-jwt';
+import jwks from 'jwks-rsa';
+
+import config from './config';
+
+export const jwtCheck = jwt({
```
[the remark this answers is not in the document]
armiantos: Good catch! Will update

www/src/providers/Auth0ProviderWithHistory.tsx
armiantos: Yes that's what I imagined, the returned scopes is an intersection of the scopes here and what the user actually has. If the user doesn't specify any scope in the beginning, we'll need to spawn another login or consent page to re-fetch the scope in the token

www/src/routes/ResumeReview.tsx
armiantos: I think it might be best to just remove it entirely and cross this bridge again when we need to. This was just for the purposes of the demo.

www/src/routes/ResumeReview.tsx
armiantos: One of two things can happen. 

* If the user is assigned a role with the specified permissions but doesn't have it in the initial token, the error will say "Consent required"
* If the user doesn't have permissions, it'll say insufficient scope

api/.env.sample
```diff
@@ -0,0 +1,4 @@
+# Taken from Auth0 API section
```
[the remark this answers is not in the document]
armiantos: Sounds good, will add 👍

www/src/routes/ResumeReview.tsx
armiantos: I'm hoping it throws a differrent exception so that we can display a different message for each

api/package.json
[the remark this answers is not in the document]
calebschoepp: IMO better to just be explicit with it. Probably never going to happen but you could imagine a future in which `jwks-rsa` no longer provides the `@types` dependency for us.

www/src/providers/Auth0ProviderWithHistory.tsx
calebschoepp: Sounds good.

www/src/routes/ResumeReview.tsx
calebschoepp: Okay, and so it seems that this will all happen client side i.e. you don't need to hit the api with invalid token to learn that it is invalid.

www/src/routes/ResumeReview.tsx
calebschoepp: Yep, let's remove it then.

api/src/routes/middleware.test.ts
```diff
@@ -1,6 +1,7 @@
 import middleware from './middleware';
 import { NextFunction, Request, Response } from 'express';
 import logger from '../util/logger';
+import * as jwtCheckModule from '../util/checkJwt';
```
calebschoepp, line 4: Nit: Verb and noun are still flipped throughout these tests. `jwtCheck` instead of `checkJwt.

api/src/routes/middleware.ts
```diff
@@ -26,7 +28,22 @@ function notFound(): Middleware {
     };
 }
 
+/**
+ * Returns middle ware that will ensure the client accessing is authenticated.
```
calebschoepp, line 32: Nit: No space in `middle ware`

calebschoepp (approved):
Couple nits but LGTM!

Ping me if you need re-approval when you push changes.

api/src/routes/middleware.ts
armiantos: Good catch (2) 😂

calebschoepp (approved):

PR #67, opened 2021-06-13
Use authenticated context

### LINKED ISSUE #58: Refactor Auth for routes
- Landing page should not have "get started" button when logged in 
- Maybe we can convert the Auth stuff to a context s.t. all components can access it

armiantos:
# Overview

Uses authenticated context to hide get started button when user is logged in. Snapshot files are automatically generated using jest. 

Enzyme is used for shallow rendering so that we can better "unit test" rather than rendering nested components

# Changes

* Hides get started button when user is authenticated
* Internal links now properly use `react-router-dom` links which doesn't require another http fetch to the web server. [Learn more about this here](https://material-ui.com/components/links/#third-party-routing-library)
* Use enzyme for UI testing

# Questions & Notes

# Issue tracking

Closes #58 

# Testing & Demo

* Check out the snapshot generated

| Logged out | Logged in |
| --- | --- |
| Screenshot_20210613_161533 | Screenshot_20210613_161554 |


www/package.json
```diff
@@ -48,12 +48,21 @@
...
         "@typescript-eslint/parser": "^4.25.0",
+        "@wojtekmaj/enzyme-adapter-react-17": "^0.6.2",
+        "enzyme": "^3.11.0",
+        "enzyme-to-json": "^3.6.2",
         "eslint": "^7.27.0",
         "eslint-plugin-prettier": "^3.4.0",
         "eslint-plugin-react": "^7.23.2",
         "prettier": "^2.3.0"
+    },
+    "jest": {
```
calebschoepp, line 63: Should this live in a `jest.config.js` file. This is the pattern we are following on the backend I believe.

www/src/routes/Landing.tsx
```diff
@@ -55,11 +57,13 @@ const Intro: FC = () => {
...
                         </Grid>
-                        <Grid item>
-                            <Button variant='contained' color='primary' onClick={() => console.log('clicked Get Started button')}>
-                                {'Get Started'}
-                            </Button>
-                        </Grid>
+                        {!isAuthenticated && (
+                            <Grid item>
+                                <Button variant='contained' color='primary' onClick={() => console.log('clicked Get Started button')}>
+                                    {'Get Started'}
```
calebschoepp, line 63: What is the point of the `Get Started` button? Why do we need to hide it when the user is logged in?

calebschoepp (commented):
Good work. Left a couple comments. I don't feel like the final authority on `www` so I won't someone else to review this too.

www/package.json
armiantos: I would like to but I think to make jest work for react, `create-react-app` has a preconfigured template and their guidance to override it is to put it in `package.json`. I tried exporting their default config but still no luck and am getting the following errors
image

www/src/routes/Landing.tsx
armiantos: I think its supposed to ask someone to log-in

www/package.json
calebschoepp: Okay that is fine then.

www/src/routes/Landing.tsx
cyrusdiego: ^ yup, the `Get Started` button is just to bring them to the Login page

cyrusdiego (approved):
This LGTM! Good work on this :)

PR #68, opened 2021-06-14
Landing page update 

cyrusdiego:
# Overview

Updated Landing page to be used for static site deployment 

# Changes

- Updated CompE+ logo 
- Updated "wave" background design 
- Updated static content (i.e. the text) 
- Commented out header links for now until #61 is merged 
- Added fade in animations
- Anchor points to more info and to mailing list sections 
# Questions & Notes

- Once this PR is merged, I'm going to deploy `www-prod` branch on netlify (domain has been purchased - compe.plus) so that we have something live and are able to share with others 
- One issue I found was when resizing the browser window, there would sometimes be a very thin space between the dark green div and the "wave" svg pattern. This only happens with the bottom div though

# Issue tracking

Closes #57 

# Testing & Demo
image
image
image
image
image
image

PR #123, opened 2021-06-22
Mobile view for static

### LINKED ISSUE #122: Make static site mobile friendly
- icons should scale appropriatley 
- copy could stack vertically rather than horizontally
- fade in should not overlap

cyrusdiego:
# Overview

Create mobile version of static site.

I wanted to keep this version as simple as possible to avoid further dev work on just layout / making it responsive. So this will be a temporary solution for users going to the site on mobile. I think when this is fully deployed, I'd want to essentially render a mobile-friendly version of our static site if a user opens it on their phone. This way, they can still get information about the program but they'd still need to go on their desktop to login and make use of the services. 

Wait until #68 is merged 

# Changes

- Imported react-device-detect to detect if the app is open on a mobile device 
- Added MobileLanding route 

# Issue tracking

Closes #122 

# Testing & Demo
To test locally (I just found this out lmao), find your IPv4 address then open your browser on your phone and enter "IP_ADDRESS:3000" with your localhost running on your computer.

image
image

### DIFF
--- a/www/package-lock.json
+++ b/www/package-lock.json
@@ -23,6 +23,7 @@
                 "@types/redux": "^3.6.0",
                 "axios": "^0.21.1",
                 "react": "^17.0.2",
+                "react-device-detect": "^1.17.0",
                 "react-dom": "^17.0.2",
                 "react-redux": "^7.2.4",
                 "react-router-dom": "^5.2.0",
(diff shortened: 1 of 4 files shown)


calebschoepp (approved):
LGTM!

PR #124, opened 2021-06-22
CI for www-prod

cyrusdiego:
# Overview

Added www-prod to github actions


armiantos (approved):
Yep this is how you do it 💯

PR #130, opened 2021-06-23
Add CI to www-prod

cyrusdiego:
# Overview

Targetted wrong branch in #124 (face-palm) 
Had to add a mock for IntersectionObserver. See [this](https://stackoverflow.com/questions/44249985/js-testing-code-that-uses-an-intersectionobserver) stackoverflow article

# Changes

Just cherry picked commit from #124 and merging to the correct target branch

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,13 @@ name: Continuous Integration
 
 on:
     push:
-        branches: [master]
+        branches:
+            - master
+            - www-prod
     pull_request:
-        branches: [master]
+        branches:
+            - master
+            - www-prod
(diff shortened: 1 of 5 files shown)


armiantos:
I tried to mock IntersectionObserver but this also looks like a viable alternative [react-intersection-observer](https://www.npmjs.com/package/react-intersection-observer)

armiantos (approved):
That is some weird function syntax but I'm okay if we just need that as a mock object for the test lol LGTM

armiantos:
And as an FYI using react-intersection-observer will still introduce the IntersectionObserver not defined

PR #132, opened 2021-06-23
Move enzyme boilerplate to `setupTests.ts`

### LINKED ISSUE #131: Avoid enzyme boilerplate
Currently all shallow calls must be preceded with an Enzyme configuration. This means setting up Enzyme for each test file.

armiantos:
# Overview

* Individual tests no longer need to set-up enzyme per test file

# Changes

* Move Enzyme configuration to `setupTests`

# Questions & Notes

# Issue tracking

Closes #131 

# Testing & Demo

Running `npm test` in `www` should not yield any errors

### DIFF
--- a/www/src/setupTests.ts
+++ b/www/src/setupTests.ts
@@ -1,7 +1,3 @@
-// jest-dom adds custom jest matchers for asserting on DOM nodes.
-// allows you to do things like:
-// expect(element).toHaveTextContent(/react/i)
-// learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
 
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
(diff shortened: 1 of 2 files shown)


www/src/setupTests.ts
```diff
@@ -3,3 +3,8 @@
 // expect(element).toHaveTextContent(/react/i)
```
calebschoepp, line 3: Nit: Are these comments necessary?

calebschoepp (approved):
LGTM!

www/src/setupTests.ts
armiantos: No, they just came with create-react-app, should we remove them?

armiantos: Given that we can always go to [jest-dom](https://github.com/testing-library/jest-dom), I'm in favor of removing this

PR #133, opened 2021-06-24
Improve authentication middleware

calebschoepp:
# Overview

Improve authentication middleware.

# Changes

- Catch error thrown by `checkJwt` when a request is not authenticated
- Add a new `NotAuthenticatedException`

# Questions & Notes

N/A

# Issue tracking

Next PR will add authentication 

# Testing & Demo

See unit tests

### DIFF
--- a/api/src/controllers/middleware.test.ts
+++ b/api/src/controllers/middleware.test.ts
@@ -98,6 +98,14 @@ describe('errorHandler middleware', () => {
         expect(resStatus).toBeCalledWith(404);
     });
 
+    it('handles authentication errors', () => {
+        const e = new Error('asdf');
+        e.name = 'UnauthorizedError';
+        middleware.errorHandler()(e, mockRequest as Request, mockResponse as Response, nextFunction);
+        expect(resJson).toBeCalledWith({ code: 401, message: 'Not authenticated' });
+        expect(resStatus).toBeCalledWith(401);
+    });
+
     // Can't figure out how to mock req.log.error so I'm punting on this
     it.todo('handles unknown errors');
 });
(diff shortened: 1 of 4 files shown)


api/src/controllers/middleware.ts
```diff
@@ -40,6 +41,8 @@ function errorHandler(): ErrMiddleware {
         if (err instanceof HttpException) {
             // Handle a known server error
             outErr = err;
+        } else if (err.name === 'UnauthorizedError') {
```
armiantos, line 44: Nit: Instead of relying on this name, we can actually use `err instance of UnauthorizedError` from `express-jwt`

```ts
import { UnauthorizedError } from 'express-jwt';
```

and I think we can update the tests as well

api/src/controllers/middleware.test.ts
```diff
@@ -98,6 +98,14 @@ describe('errorHandler middleware', () => {
         expect(resStatus).toBeCalledWith(404);
     });
 
+    it('handles authentication errors', () => {
+        const e = new Error('asdf');
+        e.name = 'UnauthorizedError';
```
armiantos, line 103:
```ts
        const e = new UnauthorizedError('credentials_required', { message: '' });
```
If we're going with `UnauthorizedError`

armiantos (commented):
One nit, let me know what you think. I'm also down for just using the string since that provides some form of decoupling with the libraries that we use

api/src/controllers/middleware.ts
calebschoepp: Omg yes this is WAY BETTER! Thanks Armi, great catch!

calebschoepp:
> One nit, let me know what you think. I'm also down for just using the string since that provides some form of decoupling with the libraries that we use

@armiantos  I like going off the type way more. Not sure why I didn't try it.

cyrusdiego (approved):
LGTM!

armiantos (approved):
LGTM!

PR #134, opened 2021-06-24
Update fluent validation library

calebschoepp:
# Overview

Update fluent validation library. There was a bug in the library that prevented chaining `mustAsync` method calls. I opened an [issue](https://github.com/AlexJPotter/fluentvalidation-ts/issues/18) and the maintainer fixed it :smile:.


armiantos (approved):
LGTM!

PR #136, opened 2021-06-26
Create redux schema for resume review

### LINKED ISSUE #96: Set-up react-redux
* Install `react-redux` and possibly `@reduxjs/toolkit`
* Add an empty store provider in `index.tsx`
* https://react-redux.js.org/tutorials/quick-start

armiantos:
# Overview

* Provide template to render different UI according to the user role
* UI state structure varies quite significantly from role to role which is why we have multiple Redux stores
  * By separating the stores, we can better guard e.g. student components from tampering the admin state (because that store doesn't exist if the current role is a student)
* The role mapping is slightly different to reduce the number of components we have to write
  * Default role -> `student`
  * `reviewer` and `interviewer` -> `volunteer`
  * `admin` -> `admin`

# Changes

* Install react-redux following [Quick Start | Redux](https://redux.js.org/tutorials/quick-start) and [Quick Start (Typescript) | Redux](https://redux.js.org/tutorials/typescript-quick-start)
* Created 4 stores:
  * One global store to hold current role
  * 3 separate stores for the 3 roles

# Questions & Notes

* React-redux and enzyme don't play very nicely yet so we can't do snapshot testing but we can still use `testing-library/react`'s `render` which actually renders out the HTML but only test for stuff that the component is directly responsible for (i.e. tests for the broadest detail possible like className or text in its direct child)

# Issue tracking

Closes #96 

# Testing & Demo

Refer to unit tests in `www`


www/src/App.tsx
```diff
@@ -17,28 +22,66 @@ const header_sections: Section[] = [
...
+                        <Landing />
+                    </Route>
+                </Switch>
+            </Router>
+        </Provider>
+    );
+};
+
+const VolunteerApp: FC = () => {
+    // TODO: Update components accordingly
```
cyrusdiego, line 50: Can you create an issue for this (if one hasn't been made already) pleaseeee

www/src/App.tsx
```diff
@@ -17,28 +22,66 @@ const header_sections: Section[] = [
...
+    return (
+        <Provider store={volunteerStore}>
+            <p>🚧 Work in progress 🚧</p>
+        </Provider>
+    );
+};
+
+const AdminApp: FC = () => {
+    // TODO: Update components accordingly
```
cyrusdiego, line 59: Same comment as above

www/src/App.tsx
```diff
@@ -17,28 +22,66 @@ const header_sections: Section[] = [
...
+    return (
+        <Provider store={adminStore}>
+            <p>🚧 Work in progress 🚧</p>
+        </Provider>
+    );
+};
+
 const App: FC = () => {
+    const currentRole = useAppSelector((state) => state.user.currentRole);
```
cyrusdiego, line 68: Might be best to wrap this role to content mapping in a function in case future roles come up or more complicated mapping comes up.

www/src/redux/hooks.ts
```diff
@@ -0,0 +1,7 @@
+import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux';
+
+import type { AppDispatch, RootState } from './store';
+
+// Use throughout your app instead of plain `useDispatch` and `useSelector`
+export const useAppDispatch = (): AppDispatch => useDispatch<AppDispatch>();
```
cyrusdiego, line 6: This is so dispatch and selector are typed right? Like this is to use `useDispatch` and `useSelector` with TS?

www/src/redux/slices/userSlice.ts
```diff
@@ -0,0 +1,12 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+export const counterSlice = createSlice({
```
cyrusdiego, line 3:
```suggestion
export const userSlice = createSlice({
```

www/src/redux/slices/userSlice.ts
```diff
@@ -0,0 +1,12 @@
...
+    name: 'user',
+    initialState: {
+        roles: [],
+        currentRole: 'student',
+    },
+    reducers: {},
+});
+
+export default counterSlice.reducer;
```
cyrusdiego, line 12:
```suggestion
export default userSlice.reducer;
```

cyrusdiego (changes requested):
This LGTM! Just one change before we merge, but overall looks great and should let us manage our code really well :D

www/src/redux/slices/userSlice.ts
armiantos: Good catch!

www/src/redux/hooks.ts
armiantos: Yes, this corresponds to the [TypeScript Quick Start | Redux](https://redux.js.org/tutorials/typescript-quick-start#define-typed-hooks) section

www/src/App.tsx
armiantos: Created #138 👍

www/src/App.tsx
armiantos: Created #139

www/src/App.tsx
armiantos: Sounds good 👍

www/src/redux/slices/userSlice.ts
armiantos: Updated!

cyrusdiego (approved):
LGTM! Good work on this :D

armiantos:
Thank you for your feedback and help during the design discussion 😄

PR #137, opened 2021-06-27
RFC: Authorization

calebschoepp:
# Overview

This PR outlines how I think we should do Authorization and the changes to our endpoints that this will require. These changes will be implemented in subsequent PRs but I still want this PR to merge so that `endpoints.md` is checked in. There is a lot to cover so I will try to step through it linearly.

We have a number of endpoints, we need to make sure that the callers to those endpoints are authorized to access them. We can use an `authorization` endpoint to make sure that the caller has the right scopes (aka permissions) when calling an endpoint. Each different role (student, reviewer, interviewer, admin) in Auth0 is associated with a number of scopes.

```typescript
router.get('/resume-reviews', middleware.authorize(scopes.ReadResumeReview), controller.getResumeReviews);
```

However, for certain endpoints we want more granular authorization. For example, with the above get resume reviews endpoint, students should only be able to read their resume reviews where as reviewers and admins should be able to read any resume review. How can we deal with this?

I propose that we would make a controller for each authorization use case with the appropriate logic. Thus for any endpoint where there are multiple possible scopes we would have a controller for each scope.

```typescript
router.get('/resume-reviews', middleware.authorizeAndFallthrough(scopes.ReadAllResumeReview), controller.getAllResumeReviews);
router.get('/resume-reviews', middleware.authorize(scopes.ReadMyResumeReview), controller.getMyResumeReviews);
```

Note two things above: First, that the more generic permission is matched for first. Second, that a new middleware `authorizeAndFallthrough` is used; this is to make sure that if the user is unauthorized for that controller it will check the next one and not just immediately return a `403` error. If we didn't have this and someone with the `read_my:resume_review` scope called this endpoint they would just always get a `403`.

This approach to authorization allows us to have fine-grained control of who can do what and limits duplicated code and extra logic. One compromise was made though.

In an idiomatic REST API you should be able to reach all resources under a url like `/resources` and a specific resource under a url like `/resources/{resource-id}`. I was following this pattern with our API. However, in cases where there were two scopes for a GET endpoint and both the general and specific cases I thought it was a bit extreme to write four controllers with nearly identical logic. As a compromise I'm going to add the ability to filter for a specific resource with a query parameter e.g. `/resources?resource_id=<your_resource_id>`


cyrusdiego (approved):
I think this makes sense to me, no questions from me. I'll give the approval but maybe wait for the others to take a look and if they have any questions / comments. Good work on this, excited to see it implemented :)

api/endpoints.md
```diff
@@ -0,0 +1,181 @@
...
+
+Create a new resume review.
+
+### Authorization
+
+-   `create:resume_reviews` — Student, Admin
+
+## `PATCH /resume-reviews/{resume-review}`
+
+Update a resume review.
```
armiantos, line 78: Can you remind me what this updates?

api/endpoints.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Query Parameters
+
+-   `?interviewer=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX`
+-   `?interviewee=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX`
+-   `?after=2021-03-21T13:08:98Z`
+
+## `POST /interviews`
+
+Create an interview.
```
armiantos, line 169: Just to clarify, this can be interepreted as `Book an interview`?

armiantos (approved):
Overall looks good to me, had a couple clarifying questions below

api/endpoints.md
calebschoepp: That is the correct interpretation.

api/endpoints.md
calebschoepp: You would use this to update the `state` of a resume review for example e.g. a resume review is finished and you need to mark it as `finished`.

PR #142, opened 2021-06-29
`www-prod` release 

cyrusdiego:
# Overview

Final "catch-all" PR for small changes to the static site prior to release 

# Changes

- Uncommented header sections 
- Not code related change but on MailChimp's side, on signup to the mailing list an automated email will be sent to the email confirming they joined the mailing list. 

# Issue tracking

List any relevant work that still needs to be done. Open and link to issues for them if applicable.

Does this close any tasks?

# Testing & Demo
`cd www && npm run start`

### DIFF
--- a/www/src/components/Header.tsx
+++ b/www/src/components/Header.tsx
@@ -1,5 +1,6 @@
 import { AppBar, Hidden, Link, makeStyles, Toolbar } from '@material-ui/core';
 import React, { FC } from 'react';
+import { Link as RouterLink } from 'react-router-dom';
 
 import logo from '../assets/logo_white.svg';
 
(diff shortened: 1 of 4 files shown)


cyrusdiego:
add compeclub website url

www/src/components/Header.tsx
```diff
@@ -28,11 +28,13 @@ export const Header: FC<HeaderProps> = (props: HeaderProps) => {
...
-                {sections.map((section) => (
-                    <Link color='textSecondary' key={section.title} variant='body2' href={section.url} className={classes.toolbarLink}>
-                        {section.title}
-                    </Link>
-                ))}
+                <Hidden xsDown>
+                    {sections.map((section) => (
+                        <Link color='textSecondary' key={section.title} variant='body2' href={section.url} className={classes.toolbarLink}>
+                            {section.title}
+                        </Link>
```
armiantos, line 35: Nit: since these are now just internal links could you add `component={RouterLink}` where `RouterLink` is imported from `react-router-dom` as 
```ts
import { Link as RouterLink } from 'react-router-dom';
```
You might also need to change `href` to `to`
For more details refer to [Header.tsx](https://github.com/UAlbertaCompEClub/compe-plus/blob/master/www/src/components/Header.tsx)
This will allow users to navigate faster between pages

You don't need to change the rest of the links since they're external

armiantos (approved):
Overall looks good to me! Thanks for doing this

The minor improvement is optional, let me know if you run into deployment issues 👍

PR #144, opened 2021-06-29
Update Postgres schema to support authorization

calebschoepp:
# Overview

Update Postgres schema to support authorization nicely

# Changes

- Make `user.id` a `TEXT` field that will hold the Auth0 user id i.e. `sub` claim on a JWT
- Update auth helper tool to use `student` role name instead of `user` role name
- Update schema diagram
- Update Zapatos schema
- Fix `npm run start`

# Testing & Demo

Ran migrations locally


armiantos (approved):
LGTM! Thanks for taking this 😄

PR #146, opened 2021-06-30
Implement authorization and update endpoints accordingly

calebschoepp:
# Overview

This PR implements authorization for our endpoints. Sorry it is so massive!

# Changes

- Make authorization middleware
- Rejig endpoints to use the middleware
- Refactor authentication middleware to follow same format as authz middleware
- Numerous other small changes

# Questions & Notes

Note I set up all of the associated scopes in our auth 0 account

# Testing & Demo

Unit tests and manual calls against the service


api/tools/urlEncode.ts
```diff
@@ -0,0 +1,3 @@
+(() => {
+    console.log(encodeURIComponent(process.argv[2]));
+})();
```
armiantos, line 3: Was the intention of this file being used using
```
npx ts-node tools/urlEncode "<str>"
```
or did you mean to include a script in `package.json`?

api/src/controllers/resumeReview/getAllResumeReviews.ts
```diff
@@ -37,14 +44,18 @@ type ResBody = {
...
  * @param res HTTP response.
- * @returns HTTP response.
+ * @returns All resume reviews.
  */
-const getResumeReviews = controller(async (req: Request<unknown, ResBody, unknown, ReqQuery>, res: Response<ResBody>): Promise<void> => {
+const getAllResumeReviews = controller(async (req: Request<unknown, ResBody, unknown, ReqQuery>, res: Response<ResBody>): Promise<void> => {
     await new ReqQueryValidator().validateAndThrow(req.query);
 
-    const allResumeReviews = await resumeReviewRepository.get(req.query.reviewee, req.query.reviewer, req.query.state);
+    const id = decodeQueryToUser(req.query.id);
```
armiantos, line 52: Nit: I think we can safely `decodeQueryToUser` because
1. Id is a uuid which is URL safe (I think, [source](https://stackoverflow.com/questions/23146193/is-a-uuid-url-safe))
2. `decodeQueryToUser` implies we want a user ID out of the query field but `id` isn't really a user ID

api/src/util/customRequest.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    gty?: string;
+    /** Permissions. */
+    permissions?: string[];
+}
+
+declare namespace Express {
+    export interface Request {
+        user: User;
+    }
+}
```
armiantos, line 26: Is the user field automatically filled in by the middleware?

armiantos (commented):
Haven't gone through the tests thoroughly but the implementation looks awesome! Had a couple questions and comments before I do another review tonight

api/tools/urlEncode.ts
calebschoepp: Yep that was the intended use

api/src/controllers/resumeReview/getAllResumeReviews.ts
calebschoepp: Yep this is a big mistake on my part. I need to change id back to being a uuid. I got mixed up when I was making all of these changes.

api/src/util/customRequest.d.ts
calebschoepp: Yes that is correct

calebschoepp:
Ready for re-review. I reordered the routes and I kept the following format:

```typescript
export { default as getAllResumeReviews } from './getAllResumeReviews';
export { default as getMyResumeReviews } from './getMyResumeReviews';
export { default as postResumeReview } from './postResumeReview';
```

B/c it broke without it. I think b/c we need to explicitly select the default export from a module.

api/src/util/customRequest.d.ts
armiantos: Sounds good, thanks 😄

api/tools/urlEncode.ts
armiantos: Sounds good, thanks for including this 😄

armiantos (approved):
LGTM! Great work 😄

PR #147, opened 2021-07-04
Routing + Netlify

cyrusdiego:
# Overview

Added `_redirects` file to properly handle react routing with netlify 

# Changes

Added `_redirects` file in `/public` directory 

# Testing & Demo

- once pushed and Netlify has built the site, test routing at www.compe.plus


armiantos (approved):
LGTM! Thanks for doing this 🍻

PR #152, opened 2021-07-08
Post user endpoint

### LINKED ISSUE #71: Implement post users endpoint
`POST /users`

calebschoepp:
# Overview

Implement endpoint to register users.

# Changes

- Implement post user endpoint
- Update Zapatos schema
- Change some other things around a necessary
- Make a new repo to hit auth 0 managment api
- New config for auth0 connection

# Questions & Notes

How do I check for error type with authorization?

# Issue tracking

Closes #71 

# Testing & Demo

Unit tests and local curling.


api/src/controllers/controllerUtil.ts
```diff
@@ -1,14 +1,14 @@
 import { NextFunction, Request, Response } from 'express';
 
-type Handler = (req: Request, res: Response, next: NextFunction) => Promise<void>;
+type Handler<P, RS, RQ, Q, L> = (req: Request<P, RS, RQ, Q, L>, res: Response<RS, L>, next: NextFunction) => Promise<void>;
```
calebschoepp, line 3: Had to do some fancy generic stuff b/c I was having trouble with typing. Happy to go into details if people are interested.

api/src/controllers/middleware.ts
```diff
@@ -100,23 +110,40 @@ function authorize(scope: Scope): AuthMiddleware {
...
-function authorizeAndFallThrough(scope: Scope): AuthMiddleware {
+function authorizeAndFallThrough(scope: Scope): ErrHandledMiddleware {
     // Check that the caller has the proper scopes
     const checkAuthorization = jwtAuthz([scope], { failWithError: true });
 
     // Catch errors thrown off by jwtAuthz and fallthrough
     const errorHandler = (err: Error, req: Request, res: Response, next: NextFunction) => {
         // Fallthrough to the next route in the router
-        next('route');
+        next('route'); // TODO only convert error when thrown by jwtAuthz
```
calebschoepp, line 120: How the heck do I check that the error was thrown only by jwtAuthz? @armiantos

armiantos: Can we do
```ts
if (err instance of <JwtErrorAuthType>)
```
I'll have to test out on my machine to find out what the error type is though. Leave this to me, I'll get it done by tonight

calebschoepp: That's what I wanted to do, but they don't use a custom error. They just return a generic object.

armiantos: Yea, I looked at [their source code](https://github.com/auth0/express-jwt-authz/blob/4591867995181a4568fb38f9c4b00c658116a815/lib/index.js#L12-L18) and it seems like they're just returning an object with 3 properties (not even an Error object). 

So I think best we can do is `if (err.message === 'Insufficient scope')` because:
* this lets us stick to our es5 `Error` type
* won't require nasty `any` castings

api/src/controllers/user/postUser.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (matches.length > 0) {
+        res.status(409).json({ user: matches[0] });
+        return;
+    }
+
+    // By default give the user the Student role in Auth0
+    await auth0Repository.giveUserRole(req.body.id, Role.Student);
+
+    // Add the user to postgres with the appropriate roles
+    const results = await userRepository.create({
```
cyrusdiego, line 74: Can this ever fail?

api/src/controllers/middleware.ts
calebschoepp: Cool, agreed. I'll make the change and then would like another review.

api/src/controllers/user/postUser.ts
calebschoepp: Theoretically yes? In all likelihood it probably won't. In any case it shouldn't be an issue if it fails b/c the only thing we already did was give the default `student` role which we want people to have anyways.

api/src/controllers/middleware.ts
calebschoepp: Just pushed the fix

api/src/controllers/controllerUtil.ts
armiantos: These are just the parameters listed in [`RequestHandler`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/df6c94b0ebb1446f1be2ccde75d8abb2340e205c/types/express/index.d.ts#L119-L124) right?

api/src/controllers/user/postUser.test.ts
```diff
@@ -0,0 +1,253 @@
...
+//     };
+//     req.body = { reviewee: '67e8ff47-8c31-4725-885c-e0e40455e7f5' };
+//     mockUserRepository.get.mockResolvedValueOnce([u]);
+
+//     await postUser(req as Request, res as Response, next);
+
+//     expect(next).not.toBeCalled();
+//     expect(res.status).toBeCalledWith(201);
+//     expect(res.json).toBeCalledWith({ resumeReview: r });
+// });
```
armiantos, line 253: I believe this is covered in 217-220?

api/src/controllers/user/postUser.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    req = {
+        body: {
+            id: 'google-oauth2|166667066666652466666',
+            email: 'test@ualberta.ca',
+            ccid: 'test',
+            program: 'Computer Engineering Software Option',
+            year: 2,
+            givenName: 'Chuck',
+            familyName: 'Norris',
+            fullName: 'Chuck Norris',
```
armiantos, line 30: nice 🍻

api/src/controllers/validation.ts
```diff
@@ -84,5 +83,23 @@ const beProperlyUriEncoded = {
...
 
-export { beAResumeReviewState, beAValidUser, beAValidUuid, beProperlyUriEncoded };
+/**
+ * Test whether a field is a valid url.
+ */
+const beAValidUrl = {
+    predicate: async (field: string | undefined): Promise<boolean> => {
+        if (field === undefined) {
+            return false;
+        }
```
armiantos, line 93: Nit: we can collapse this to the try catch, `new URL(undefined)` fails
image

armiantos (changes requested):
Overall really great! Implementation matched standards (e.g. Auth0 management calls [ref](https://auth0.com/docs/users/assign-roles-to-users)).

Additional stuff that was covered (big thank you 🙏):
* Added missing `beProperlyUriEncoded` tests

Man props to you and your error checks hahahah I've never seen such rigorous null, empty, valid checks

api/src/controllers/controllerUtil.ts
calebschoepp: That is correct

api/src/controllers/user/postUser.ts
calebschoepp: Nit: Actually a `500` but yes we would know it failed.

api/src/controllers/user/postUser.test.ts
calebschoepp: LOLZ. Totally need to delete that. That was copy-pasta from another file when I was starting implementation.

api/src/controllers/validation.ts
calebschoepp: Good catch but going to keep it as is. Without the if statement the compiler complains about assigning `string?` to `string` and I would need to do something like `field!`

armiantos: Ah good call, sounds good!

api/src/controllers/controllerUtil.ts
armiantos: Sounds good 😄

armiantos (approved):
LGTM! Awesome job 💯

calebschoepp:
> LGTM! Awesome job

Thanks for all your gracious compliments. This was a fun one.

PR #157, opened 2021-07-15
Create user registration service

### LINKED ISSUE #156: Create registration service

armiantos:
# Overview

* Provides the service call for the UI component to forward userInformation as a registration request to the backend

Usage in component will look something like 

```ts
    const { user } = useAuth0();
    const dispatch = useAppDispatch();

    dispatch(
        registerUser({
            userInfo: {
                ccid: 'sumitro',
                email: user?.email ?? '',
                familyName: 'Sumitro',
                fullName: 'Armianto Sumitro',
                givenName: 'Armianto',
                id: user?.sub ?? '',
                program: 'Software Engineering',
                year: 5,
            },
            tokenAcquirer: getTokenSilently(),
        }),
    );
```

where the userInformation object itself is fetched from the redux store. Although the thunk could be hacked to get access to the store directly, it's more functional this way.

# Changes

* Created a `POST` wrapper for `axios` and `auth0`
* Fix missing `/api/v1/` prefix on back-end requests from #168 (sorry ☹️)
* Rename `userMe` endpoint to just `me`

# Questions & Notes

* Will have to update if #171 gets merged first

# Issue tracking

Closes #156 

# Testing & Demo

Refer to unit tests

### DIFF
--- a/www/src/util/serverResponses.d.ts
+++ b/www/src/util/serverResponses.d.ts
@@ -12,4 +12,6 @@ export type User = {
     updated_at: db.TimestampTzString;
 };
 
-// TODO: Find better way of syncing server reponse types
+export type WrappedUser = {
+    user: User;
+};
--- a/www/src/util/serverResponses.d.ts
+++ b/www/src/util/serverResponses.d.ts
@@ -1,0 +1,15 @@
+export type User = {
+    id: string;
+    email: string;
+    ccid: string;
+    program: string;
+    year: number;
+    given_name: string;
+    family_name: string;
+    full_name: string;
+    photo_url: string | null;
+    created_at: db.TimestampTzString;
+    updated_at: db.TimestampTzString;
+};
+
+// TODO: Find better way of syncing server reponse types
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # CompE+
 
-TODO
+TODO: Like we should really do this.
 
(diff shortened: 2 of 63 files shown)


www/src/util/serverResponses.d.ts
```diff
@@ -12,4 +12,6 @@ export type User = {
     updated_at: db.TimestampTzString;
```
calebschoepp, line 12: Can you help me understand what is going on in this file. Why do we need to wrap this type?

calebschoepp (commented):
Nice work. Overall looks good but I'm rough on my frontend stuff rn so maybe get Cyrus to look at it too.

www/src/util/serverResponses.d.ts
armiantos: The server response JSON looks like 
```json
{
  "user" : { ... }
}
```
according to [`getMe.ts`](https://github.com/UAlbertaCompEClub/compe-plus/blob/35bbc9d6d2fd28bf45a0beae3316d28d0ec29a84/api/src/controllers/user/getMe.ts#L7-L9).

It still works fine without since we're just checking if the response returns any data at all but might be better to actually type what the server actually responds with

cyrusdiego (approved):
This LGTM, no questions on my end and good work :) sorry for the delay in the review haha

PR #166, opened 2021-07-21
Add `users/me` endpoint

### LINKED ISSUE #165: Add `user/me` endpoint
* Users should be redirected to the register page if they aren't registered on the back-end

armiantos:
# Overview

* Will be used in the front end to determine whether the user should be redirected to the register page

# Changes

* Add **GET** `/user/me` endpoint

# Questions & Notes

N/A

# Issue tracking

Closes #165 

# Testing & Demo

Refer to unit tests or curl `/users/{userId}` directly using a valid JWT token (you can fetch this from the front end application we have and make sure the `userId` path parameter is URI encoded)


api/src/controllers/user/getUser.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type ResBody = {
+    exists: boolean;
+};
+
+/**
+ * Get my resume reviews.
+ * @param req HTTP request.
+ * @param res HTTP response.
+ * @returns Your resume reviews.
```
calebschoepp, line 29: This needs to be updated from the copy/pasta

api/src/controllers/user/getUser.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+const getUser = controller(async (req: Request<ReqParams, ResBody, unknown, unknown>, res: Response<ResBody>): Promise<void> => {
```
calebschoepp, line 31: You can just stop adding generics once you don't want to change them anymore. In other words `Request<ReqParams, ResBody>`. You would need those unknowns if you were doing something like `Request<ReqParams, ResBody, unknown, unknown, AnotherType>`

api/src/controllers/user/getUser.ts
```diff
@@ -0,0 +1,46 @@
...
+    const requested = decodeQueryToUser(req.params.user);
+    const requester = req.user.sub;
+
+    if (requester !== requested) {
+        throw new NotAuthorizedException();
+    }
+
+    const matches = await userRepository.get(requested);
+
+    res.status(200).json({ exists: matches.length == 1 });
```
calebschoepp, line 43: Like you suggested I would probably return the user object.

If the user is not registered I might return a status code 404 but a null user object also works.

I also like the idea of `/users/me`. The `{user}` doesn't add much.

calebschoepp (changes requested):
Great work. Hope it was fairly clear where you needed to make changes. Left a couple comments.

api/src/controllers/user/getUser.ts
armiantos: Ah gotcha, thanks for letting me know. Will update 👍

api/src/controllers/user/getUser.ts
armiantos: Agreed. I'm not sure how axios handles 404 but I do like that it better represents `NotFoundError` so let's go with 404 for now and update if necessary

calebschoepp (approved):
LGTM! 🚢

PR #170, opened 2021-07-27
Refactor EditRolesDialog to use Redux state

### LINKED ISSUE #169: Integrate settings modal with redux

armiantos:
# Overview

* Give a practical example of how to refactor useState to redux state

# Changes

* Introduce `isEditRolesDialogOpen` to `userSlice`
* Use `setCurrentRole`in `EditRolesDialog`

# Questions & Notes

* I'll follow up this PR to clean up some of the UI component (use `roles.map(role => ...)`, rename to `SettingsDialog.tsx`, etc.)
* We need to fetch roles from server in a future PR

# Issue tracking

Closes #169

# Testing & Demo

Refer to unit tests

Screen Shot 2021-07-26 at 9 54 04 PM 1

### DIFF
--- a/api/src/controllers/document/index.ts
+++ b/api/src/controllers/document/index.ts
@@ -1,1 +1,3 @@
+export { default as patchAllDocument } from './patchAllDocument';
+export { default as patchMyDocument } from './patchMyDocument';
 export { default as postDocument } from './postDocument';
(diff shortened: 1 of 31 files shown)


cyrusdiego (approved):
LGTM! The one thing I realize though is we'd need to hide or completely disable this dialog if the user is only a student as they wouldn't have any other views to access. I'm gonna log that as an issue but shouldn't be too bad.

armiantos:
Thanks! Sounds good, yea it shouldn't be too bad to disable the dialog

PR #172, opened 2021-07-27
Get document

### LINKED ISSUE #78: Implement get documents endpoint
`GET /resume-reviews/{resume-review}/documents?is_review=true&state=submitted`

### LINKED ISSUE #80: Implement get document endpoint
`GET /resume-reviews/{resume-review}/documents/{document}`

calebschoepp:
# Overview

Implement get document endpoints.

# Changes

- Implement get my document endpoint
- Implement get all document endpoint

# Issue tracking

Closes #80 
Closes #78 

# Testing & Demo

```bash
curl localhost:1337/api/v1/resume-reviews/5993d292-97b4-4b8a-92d5-4936e4653e5b/documents?id=71b613ee-5bfa-4197-9eba-b362a4e5db8c&isReview=true -H "$(npx ts-node tools/getAuth.ts reviewer)" -v -s | jq
```


api/src/controllers/document/postDocument.ts
```diff
@@ -34,7 +35,7 @@ class ReqBodyValidator extends Validator<ReqBody> {
 
         this.ruleFor('isReview').notNull();
 
-        this.ruleFor('userId').mustAsync(beAValidUser);
+        this.ruleFor('userId').mustAsync(beAValidUser); // TODO this should match the user creating it
```
armiantos, line 38: The user that created the `resumeReview` object?

www/README.md
```diff
@@ -54,3 +54,11 @@ You don’t have to ever use `eject`. The curated feature set is suitable for sm
 You can learn more in the [Create React App documentation](https://facebook.github.io/create-react-app/docs/getting-started).
 
 To learn React, check out the [React documentation](https://reactjs.org/).
+
+# Dev
```
armiantos, line 58: Isn't this staging?

armiantos (approved):
LGTM! 🍻

I'm guessing the 6 download limit is on our S3 cap but I'm not worried about that since the user will probably only request 3 at most (if we decide to cache but I'm guessing we'll just fetch the backend on demand).

Awesome work! Could you remind me what's the difference between the `resumeReview` path parameter vs the `resumeReviewId` query parameter?

api/src/controllers/document/postDocument.ts
calebschoepp: Yeah good point. Not sure that I even want to add this restriction in though. This makes it impossible for the Admin to go in and mess around with things. Just put it down as a quick idea.

www/README.md
calebschoepp: Touche. I'll fix that

api/src/controllers/document/postDocument.ts
armiantos: Sounds good, we can put it aside for now I don't expect anyone to go out and take a token to override someone else's resume review. Also these are creates which aren't as destructive as deletes so, we should be fine

calebschoepp:
> I'm guessing the 6 download limit is on our S3 cap but I'm not worried about that since the user will probably only request 3 at most (if we decide to cache but I'm guessing we'll just fetch the backend on demand).

Nope not related to any S3 cap. I purely added a cap b/c I didn't want someone to be able to DDOS our system by requesting every document and then all of a sudden were downloading 100's of MBs from S3 on a single HTTP request. I chose 6 b/c I couldn't imagine a single resume review having more than 6 documents and if it consistently will then we can just up the limit.

> Awesome work! Could you remind me what's the difference between the `resumeReview` path parameter vs the `resumeReviewId` query parameter?

Wow I'm dumb haha :facepalm: . I just realised I need to be filtering the call to `documentRepository.get` with `params.resumeReview` and that the query parameter for resume review id is redundant.

armiantos:
> agine a single resume review having more than 6 documents and if it consistently will then we can just up the limit.
> 
> > Awesome work! Could you remind me what's the difference between the `resumeReview` path parameter vs the `resumeReviewId` query parameter?
> 
> Wow I'm dumb haha 🤦 . I just realised I need to be filtering the call to `documentRepository.get` with `params.resumeReview` and that the query parameter for resume review id is redundant.

lol do we want to fix that in a follow up PR?

calebschoepp:
> > agine a single resume review having more than 6 documents and if it consistently will then we can just up the limit.
> > > Awesome work! Could you remind me what's the difference between the `resumeReview` path parameter vs the `resumeReviewId` query parameter?
> > 
> > 
> > Wow I'm dumb haha facepalm . I just realised I need to be filtering the call to `documentRepository.get` with `params.resumeReview` and that the query parameter for resume review id is redundant.
> 
> lol do we want to fix that in a follow up PR?

Nah we can just fix it here. I should be good for a re-review.

armiantos (approved):
LGTM! (good catch on the parameter ordering, I missed that lol) 💯

PR #180, opened 2021-08-08
Add create resume review thunk

armiantos:
# Overview

Allows front end component to easily post the resume review data to the server as

```ts
    const dispatch = useStudentDispatch();

    dispatch(
        initiateResumeReview({
            userId: user?.sub ?? '',
            tokenAcquirer: getAccessTokenSilently,
            base64Contents: 'base64contentx',
        }),
    );
```

# Changes

* Add service call in non-exported `initiateResumeReview`
* Add thunk wrapped call in default export `initiateResumeReview`
* Prefix endpoints with `get` and `post` accordingly
* Delegate try catch for `postWithToken` and `fetchWithToken` to user
* Copy scopes from server
* Replace `db.TimestampTzString` to simply `string`

# Questions & Notes

* Might want to revisit to clean up the redux types
* Sadly `expect(async () => ...).toThrow(...)` [doesn't work](https://github.com/facebook/jest/issues/1700) and best syntax I could find looks like [`await expect(check).rejects.toThrow('error string')`](https://github.com/facebook/jest/issues/1700#issuecomment-386019524)

# Issue tracking

TODO

# Testing & Demo

Refer to unit tests

### DIFF
--- a/www/src/redux/substores/student/thunks/initiateResumeReview.test.tsx
+++ b/www/src/redux/substores/student/thunks/initiateResumeReview.test.tsx
@@ -1,4 +1,3 @@
-import { BaseThunkAPI } from '@reduxjs/toolkit/dist/createAsyncThunk';
 import { AxiosResponse } from 'axios';
 import { mocked } from 'ts-jest/utils';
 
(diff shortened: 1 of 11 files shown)


cyrusdiego (approved):
LGTM!

PR #184, opened 2021-08-12
Implement patch resume review endpoints

### LINKED ISSUE #77: Implement patch resume review endpoint
`PATCH /resume-reviews/{resume-review}`

calebschoepp:
# Overview

Implement patch resume review endpoints

# Issue tracking

Closes #77

### DIFF
--- a/api/src/controllers/resumeReview/patchAllResumeReview.ts
+++ b/api/src/controllers/resumeReview/patchAllResumeReview.ts
@@ -1,0 +1,64 @@
+import { Request, Response } from 'express';
+import type * as s from 'zapatos/schema';
+
+import InternalServerErrorException from '../../exceptions/InternalServerErrorException';
+import * as resumeReviewRepository from '../../repositories/resumeReviewRepository';
+import controller from '../controllerUtil';
+import Validator, { beAResumeReviewState, beAValidResumeReview, beAValidUser, beAValidUuid } from '../validation';
+
+type Params = {
+    resumeReview: string;
+};
+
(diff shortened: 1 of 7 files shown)


api/src/controllers/resumeReview/patchAllResumeReview.ts
```diff
@@ -0,0 +1,64 @@
...
+    state?: s.resume_review_state;
+};
+
+class ReqBodyValidator extends Validator<ReqBody> {
+    constructor() {
+        super('message body');
+
+        this.ruleFor('reviewee')
+            .mustAsync(beAValidUser)
+            .when((reqBody) => reqBody.reviewee !== undefined);
```
armiantos, line 33: Is this preferred over

```ts
        this.ruleFor('reviewee').notNull().notEmpty();
```
?

armiantos (commented):
Great work! Just had one question before approving

api/src/controllers/resumeReview/patchAllResumeReview.ts
calebschoepp: @armiantos Does something entirely different actually. Forces the validation to only run when the field is defined. If we didn't include this the validation would fail if you don't include a `reviewee` field which is something we don't want because it is possible that they don't want to update the reviewee.

[Doc](https://fluentvalidation-ts.alexpotter.dev/docs/api/configuration/when)

armiantos: Gotcha! Thanks for the clarification

PR #193, opened 2021-08-19
Hotfix for postgres connection

calebschoepp:
Turns out staging and prod could never actually connect to the database. LMAO. Here is a hotfix.


armiantos (approved):
LGTM! I don't see the`USE_SSL` variable on heroku, do we need to turn this on?

calebschoepp:
> LGTM! I don't see the`USE_SSL` variable on heroku, do we need to turn this on?

This is a custom environment variable we need to set ourselves.

PR #198, opened 2021-08-21
Logic skeleton of reviewer page

calebschoepp:
# Overview

Setup call to API for volunteer resume outline page.

# Changes

- Setup two calls to the API
- Mock UI. Will fix in followup PR

# Testing & Demo

Just ran it and saw that the network calls were being made.


www/src/redux/slices/userSlice.ts
```diff
@@ -47,7 +47,7 @@ export const userSlice = createSlice({
         builder.addCase(getUserRole.fulfilled, (state, action) => {
             state.isLoading = false;
             state.roles = action.payload?.roles.map((roleObject) => roleObject.role) ?? [];
-            state.currentRole = 'student';
+            state.currentRole = 'reviewer';
```
armiantos, line 50: Should we revert this?

armiantos (approved):
LGTM! Just had one question regarding the default role.

I can also integrate the revert to my next PR which builds on top off @rachel-ellis's and @AnushkaKhare-Eng's PR for edit roles dialog

www/src/redux/slices/userSlice.ts
calebschoepp: Good catch. I will revert this.

PR #199, opened 2021-08-22
Use https for local front end dev

armiantos:
# Overview

* Allows devs to fetch token with scopes without user consent (mimics how staging and prod behaves)

# Changes

* Use https port to serve local front end
* Add instructions on how to add new non-localhost mapping to `/etc/hosts`

# Questions & Notes

* Auth0 doesn't allow [skipping user consent for apps hosted in localhost](https://auth0.com/docs/authorization/user-consent-and-third-party-applications#skip-consent-for-first-party-applications). A dev fix for this would require us to make localhost behave as if it was hosted elsewhere
* Adding a new entry in `/etc/hosts` allows our machine to "host" on a domain other than "localhost"
* Actual non-localhost URL isn't added to prevent anyone from cloning this repository and messing up our dev environment (Auth0's original intention)

# Issue tracking

# Testing & Demo

Token fetches without prior consent from developers should no longer fail

### DIFF
--- a/www/README.md
+++ b/www/README.md
@@ -5,9 +5,9 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 ## Authentication set-up
 
 1. Copy `www/.env.sample` to `www/.env`
-2. Using the `compe-plus-dev` tenant in ![Auth0](https://manage.auth0.com/dashboard), expand _Applications_ on the left sidebar and click on _Compe+ Web Application_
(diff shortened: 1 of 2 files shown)


calebschoepp (changes requested):
Don't you need to add `HTTPS=true` to the react script?

armiantos:
> Don't you need to add `HTTPS=true` to the react script?

Good catch, I reset the commit because I didn't want to include the local domain that we decided on and forgot to include that back

calebschoepp (approved):
LGTM! Thanks for pair programming with me

PR #204, opened 2021-08-23
Fix POST document

### LINKED ISSUE #203: Initiate resume review doesn't work for students
- Resume review route param is missing
- Post document will fail with the following response if the resume file is too large
```
    req: {
      "id": 15,
      "method": "POST",
      "url": "/api/v1/resume-reviews//documents"
    }
    stackTrace: "PayloadTooLargeError: request entity too large\n    at readStream (/Users/armiantos/code/compe-plus/api/node_modules/raw-body/index.js:155:17)\n    at getRawBody (/Users/armiantos/code/compe-plus/api/node_modules/raw-body/index.js:108:12)\n    at read (/Users/armiantos/code/compe-plus/api/node_modules/body-parser/lib/read.js:77:3)\n    at jsonParser (/Users/armiantos/code/compe-plus/api/node_modules/body-parser/lib/types/json.js:135:5)\n    at Layer.handle [as handle_request] (/Users/armiantos/code/compe-plus/api/node_modules/express/lib/router/layer.js:95:5)\n    at trim_prefix (/Users/armiantos/code/compe-plus/api/node_modules/express/lib/router/index.js:317:13)\n    at /Users/armiantos/code/compe-plus/api/node_modules/express/lib/router/index.js:284:7\n    at Function.process_params (/Users/armiantos/code/compe-plus/api/node_modules/express/lib/router/index.js:335:12)\n    at next (/Users/armiantos/code/compe-plus/api/node_modules/express/lib/router/index.js:275:10)\n    at cors (/Users/armiantos/code/compe-plus/api/node_modules/cors/lib/index.js:188:7)"
    error: {
      "message": "request entity too large",
      "expected": 3738503,
      "length": 3738503,
      "limit": 3145728,
      "type": "entity.too.large"
    }
```

armiantos:
# Overview

* Fix initiating a resume review as a student

# Changes

* Wrap server responses for POST resume-review and POST document (was previously broken)
* Add env variables for size caps (both front and back ends)
* Display warning to user if file size is too large

# Questions & Notes

* Because we're base64-ing the resume contents, its file size is ~33% larger than its array buffer counter part, since express' default json middleware handles upto 100kb, this caps our resume at 100/1.33 = 75kb

# Issue tracking

Closes #203 

# Testing & Demo

Run the full stack locally without the bucketeer env variables, POST resume review should pass and POST document should fail with the following message in the backend

```
    req: {
      "id": 18,
      "method": "POST",
      "url": "/api/v1/resume-reviews/e9c0e882-51b2-45c5-bb43-fd86c723ee75/documents"
    }
    error: {
      "message": "The AWS Access Key Id you provided does not exist in our records.",
      "code": "InvalidAccessKeyId",
      "region": null,
      "time": "2021-08-23T20:33:14.413Z",
      "requestId": "D2S9FQJ5T0C5QPTZ",
      "extendedRequestId": "KgVKqOpAtnrTzhaQjk9sDqgdCGefhunxSnqLDduDZo3edZVNgqVnopqEHXkg5lFhxuvEGcFK2oU=",
      "statusCode": 403,
      "retryable": false,
      "retryDelay": 136.7333900498528
    }
```

(instead of the previous error)


api/src/util/config.ts
```diff
@@ -23,6 +23,7 @@ const BUCKETEER_AWS_REGION = process.env.BUCKETEER_AWS_REGION || 'us-east-1';
 const BUCKETEER_BUCKET_NAME = process.env.BUCKETEER_BUCKET_NAME || 'documents';
 const BUCKETEER_ENDPOINT = process.env.BUCKETEER_ENDPOINT || '';
 const USE_SSL = process.env.USE_SSL || false;
+const MAX_JSON_SIZE = process.env.MAX_JSON_SIZE || '100kb';
```
calebschoepp, line 26: I think we need to support more than 100kb. I do not think that is big enough. People will get frustrated if there resume is too big and they can't submit it. I would say we should support on the order of MBs

calebschoepp (changes requested):
Code LGTM but I think we need to increase the size limit.

api/src/util/config.ts
armiantos: Sounds good, bumping up the default values

PR #206, opened 2021-08-25
GET All users and their roles

cyrusdiego:
# Overview

Implement endpoint to get a list of all users and their roles. This will be used in the Admin view to set the roles for users. 

# Changes

- Created admin related routes, repository, and controller 

# Questions & Notes

How do we ensure only an Admin user can access this endpoint? 
- I added a new scope in Auth0 →  `read:all_users_roles` (lemme know if I can rename it to something better) and I added this to the middleware for the route; is this all that's necessary? 

# Testing & Demo

See unit tests

### DIFF
--- a/api/src/controllers/admin/getAllUsersInfo.ts
+++ b/api/src/controllers/admin/getAllUsersInfo.ts
@@ -1,0 +1,23 @@
+import { Request, Response } from 'express';
+import { CamelCasedProperties } from 'type-fest';
+
+import * as adminRepository from '../../repositories/adminRepository';
+import { userRoleSelectable } from '../../repositories/adminRepository';
+import { toCamelCase } from '../../util/helper';
+import controller from '../controllerUtil';
+
+type ResBody = { users: CamelCasedProperties<userRoleSelectable>[] };
+
+/**
+ * Get all users and their roles.
+ * @param req HTTP request.
+ * @param res HTTP response.
+ * @returns List of all users and their roles.
+ */
+const getAllUsersInfo = controller(async (_req: Request, res: Response<ResBody>): Promise<void> => {
+    const users = await adminRepository.getUsersInfo();
+
+    res.status(200).json({ users: users.map((user) => toCamelCase(user)) });
+});
+
+export default getAllUsersInfo;
--- a/api/src/repositories/adminRepository.ts
+++ b/api/src/repositories/adminRepository.ts
@@ -1,0 +1,24 @@
+import * as db from 'zapatos/db';
+import { conditions as dc } from 'zapatos/db';
+import type * as s from 'zapatos/schema';
+import { role_type } from 'zapatos/schema';
+
+import pool from '../util/pool';
+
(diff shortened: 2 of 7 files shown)


api/src/repositories/adminRepository.ts
```diff
@@ -0,0 +1,24 @@
...
+import pool from '../util/pool';
+
+type userRoleSQL = s.users.SQL | s.user_roles.SQL;
+export type userRoleSelectable = s.users.JSONSelectable & { user_role: role_type };
+
+/**
+ * Gets all users and their roles.
+ * @returns List of all users and their assigned roles.
+ */
+const getUsersInfo = async (): Promise<userRoleSelectable[]> => {
```
cyrusdiego, line 15: Definitely need a double check on this (whether or not I'm doing the join correctly)

calebschoepp: Seems right to me. Is it working and giving you the results you expect?

api/src/routes/adminRoutes.ts
```diff
@@ -0,0 +1,11 @@
+import express from 'express';
+
+import * as controller from '../controllers/admin/index';
+import middleware from '../controllers/middleware';
+import Scope from '../types/scopes';
+
+export const router = express.Router();
+
+router.get('/admin/allUsers', middleware.authorize(Scope.ReadAllUsersRoles), controller.getAllUsersInfo);
```
calebschoepp, line 9: Yep this is all you should need for authorization. Make sure you add the role to all the environments when this merges.

api/src/types/scopes.ts
```diff
@@ -23,6 +23,7 @@ enum Scope {
     ReadAllInterviews = 'read_all:interviews',
     CreateInterviews = 'create:interviews',
     DeleteInterviews = 'delete:interviews',
+    ReadAllUsersRoles = 'read:all_users_roles',
```
calebschoepp, line 26: Auth0 config should be updated to. Ping if you need help with that.

api/src/controllers/admin/getAllUsersInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Get all users and their roles.
+ * @param req HTTP request.
+ * @param res HTTP response.
+ * @returns List of all users and their roles.
+ */
+const getAllUsersInfo = controller(async (_req: Request, res: Response<ResBody>): Promise<void> => {
+    const users = await adminRepository.getUsersInfo();
+
+    res.status(200).json({ users: users.map((user) => toCamelCase(user)) });
```
calebschoepp, line 20: You can just use the `manyToCamelCase` helper.

calebschoepp (commented):
Nice work. Let me know if the join is working as expected cause I've been struggling with writing a join haha.

api/src/repositories/adminRepository.ts
calebschoepp: Actually don't you need the role in the SELECT clause?

PR #214, opened 2021-09-02
Delete Role from User 

cyrusdiego:
# Overview

Create delete role endpoint 

# Changes

Created related routes, controller, and repository.

# Testing & Demo

See unit test


api/src/routes/adminRoutes.ts
```diff
@@ -8,4 +8,6 @@ export const router = express.Router();
 
 router.get('/admin/allUsers', middleware.authorize(Scope.ReadAllUsersRoles), controller.getAllUsersInfo);
 
+router.delete('/admin/userRole', middleware.authorize(Scope.DeleteUsersRole), controller.deleteRole);
```
armiantos, line 11: Shouldn't this be in `/users/{user}/roles/{role}`?

api/src/types/scopes.ts
```diff
@@ -24,6 +24,7 @@ enum Scope {
     CreateInterviews = 'create:interviews',
     DeleteInterviews = 'delete:interviews',
     ReadAllUsersRoles = 'read:all_users_roles',
+    DeleteUsersRole = 'delete:users_role',
```
armiantos, line 27: We also already have the `delete:roles` scope for admins

cyrusdiego: Removed this line and replaced usage with the `delete:roles` scope

armiantos: Push changes?

api/src/routes/adminRoutes.ts
cyrusdiego: Hmm maybe? I'm not really sure the proper structure for this. My rationale for putting this under admin is that only admin can delete a role. But since it's properly scoped it shouldn't matter and would probably be best under `/users`. I'm not totally sure tbh, if you think I should put it under that endpoint then I can change it

api/src/types/scopes.ts
cyrusdiego: *face palm* lol pushed

api/src/routes/adminRoutes.ts
armiantos: It is more RESTFUL if we put it in `/users/{user}/roles/{role}` and it does follow `endpoints.md` but given the time constraint I'm also OK putting that here

api/src/controllers/admin/deleteRole.ts
```diff
@@ -0,0 +1,51 @@
...
+
+type ResBody = {
+    users?: CamelCasedProperties<s.user_roles.JSONSelectable>[];
+    message?: string;
+};
+
+type Params = {
+    userId: string;
+    role: string;
+};
```
armiantos, line 20: `Params` are usually used for route parameters, e.g. `users/{userId}/roles/{role}` did you mean to use `ReqBody`?

api/src/controllers/admin/deleteRole.ts
```diff
@@ -0,0 +1,51 @@
...
+        super('route parameters');
+
+        this.ruleFor('userId').mustAsync(beAValidUser);
+
+        this.ruleFor('role').mustAsync(beAValidRole);
+    }
+}
+
+const deleteRole = controller(async (req: Request<Params, ResBody>, res: Response<ResBody>): Promise<void> => {
+    await new ParamsValidator().validateAndThrow(req.params);
```
armiantos, line 33: `req.params` return the route parameters, from the route in `adminRoutes`, I don't see `userId` or `role` anywhere, did you mean to use `req.body`?

api/src/routes/adminRoutes.ts
cyrusdiego: Decided to just put it under the endpoint specified, rather do it right now than fix it later haha

api/src/controllers/admin/deleteRole.ts
armiantos: Since the route is now updated to `/users/{userId}/role/{role}` we now have route parameters. Sweet1

api/src/controllers/roles/deleteRole.ts
```diff
@@ -0,0 +1,51 @@
...
+    const indexOfAssignedRole = assignedRoles.findIndex((assignedRole) => assignedRole.role === role);
+    if (indexOfAssignedRole < 0) {
+        res.status(409).json({ message: 'User with id: ' + userId + ' did not have role: ' + role });
+        return;
+    }
+
+    await auth0Repository.removeUserRole(userId, role);
+    const result = await userRolesRepository.remove({ user_id: userId, role: role });
+
+    res.status(204).json({ users: manyToCamelCase(result) });
```
armiantos, line 48:
```suggestion
    res.status(204).end();
```

Status code 204 stands for no content (my bad I did this in my own PR, I'll go and fix that real quick hahahah)

armiantos (changes requested):
LGTM! Just one last change please 😂

api/src/controllers/roles/deleteRole.ts
cyrusdiego: oop good catch.

armiantos (approved):
LGTM! Awesome work 😄

PR #220, opened 2021-09-03
Write SQL migrations for new Calendly resource

calebschoepp:
# Changes

- New db migration to add calendly resource
- Zapatos schema updated

# Notes

I need to run these migrations on staging and prod when this merges

### DIFF
--- a/api/db/migrations/003_add_calendly_resource.up.sql
+++ b/api/db/migrations/003_add_calendly_resource.up.sql
@@ -2,7 +2,7 @@
 CREATE TABLE IF NOT EXISTS calendlys (
     id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
     link TEXT NOT NULL,
-    interviewee TEXT REFERENCES users ON DELETE RESTRICT,
+    interviewees TEXT[],
     interviewer TEXT NOT NULL REFERENCES users ON DELETE RESTRICT,
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
--- a/api/db/migrations/003_add_calendly_resource.up.sql
+++ b/api/db/migrations/003_add_calendly_resource.up.sql
@@ -1,0 +1,14 @@
+/* calendlys table */
+CREATE TABLE IF NOT EXISTS calendlys (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    link TEXT NOT NULL,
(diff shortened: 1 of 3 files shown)


armiantos:
From what I understand, the links are attached to interviewers and interviewees will be matched to those links right? Is there a reason why we'd go with this schema instead of 

```sql
CREATE TABLE calendly_links (
interviewer TEXT NOT NULL PRIMARY KEY REFERENCES users,
link TEXT NOT NULL
created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
);

CREATE TABLE calendly_interviews (
id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
interviewee TEXT REFERENCES users ON DELETE RESTRICT,
interviewer TEXT NOT NULL REFERENCES users ON DELETE RESTRICT,
);
```

I imagine we'd want to insert the volunteer link to the database when volunteers enter their calendly links, and then generate interview objects that link interviewee and interviewer when interviewees are matched.

With the current schema if we leave interviewee blank, don't we have to copy a row only to rematch the interviewer to another interviewee?

armiantos (commented):
Refer to comment above

api/db/migrations/003_add_calendly_resource.up.sql
```diff
@@ -0,0 +1,14 @@
+/* calendlys table */
+CREATE TABLE IF NOT EXISTS calendlys (
+    id UUID NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
+    link TEXT NOT NULL,
+    interviewee TEXT REFERENCES users ON DELETE RESTRICT,
```
armiantos, line 5: I think we want the following to support the one (interviewer) to many (interviewee) relationship
```suggestion
    interviewee TEXT[] REFERENCES users ON DELETE RESTRICT,
```

armiantos (changes requested):

api/db/migrations/003_add_calendly_resource.up.sql
calebschoepp: I've added this change @armiantos . Ready for rereview.

armiantos (approved):
LGTM! We'll lose the referential integrity but given that we'll be using the db stricly using the backend this should be enough. Great work!🚀

calebschoepp:
> LGTM! We'll lose the referential integrity but given that we'll be using the db stricly using the backend this should be enough. Great work!

Yeah I was a bit sad about that, but this felt like the quickest approach. We can always refactor later if necessary :D.

PR #202, opened 2021-08-23
Implement PUT role

armiantos:
# Overview

* Allows admins/devs to assign roles to users via curl

# Changes

* Add `PUT /api/v1/user/:userId/roles/:role` endpoint
* Move role routes to a separate file

# Questions & Notes

* Duplicate assignment is handled

# Issue tracking

Closes #73 

# Testing & Demo

Run the following curl

```sh
curl http://localhost:1337/api/v1/users/google-oauth2%7C100464926847930484423/roles/reviewer -X PUT  -H "$(npx ts-node tools/getAuth.ts admin)"
```

PR #221, opened 2021-09-03
Write calendly endpoints

calebschoepp:
# Overview

Endpoints to interact with calendly resource. Note that to speed up the implementation I didn't differentiate between my/any endpoints. Nothing particularly confidential about it anyways.

# Changes

- POST endpoint
- GET endpoint
- PATCH endpoint

# Questions & Notes

- Next PR is going to add authorization.
- I skipped testing for speed of implementation
- Dependent on schema migration PR #220

### DIFF
--- a/api/src/repositories/calendlyRepository.ts
+++ b/api/src/repositories/calendlyRepository.ts
@@ -1,0 +1,16 @@
+import * as db from 'zapatos/db';
+import type * as s from 'zapatos/schema';
+
+import pool from '../util/pool';
+
+/**
+ * Create a new calendly.
+ * @param link Calendly link.
+ * @param interviewer Interviewer whose calendly this is.
+ * @returns The newly created calendly.
+ */
+const create = async (link: string, interviewer: string): Promise<s.calendlys.JSONSelectable> => {
+    return db.insert('calendlys', { link, interviewer }).run(pool);
+};
+
+export { create };
--- a/api/src/repositories/calendlyRepository.ts
+++ b/api/src/repositories/calendlyRepository.ts
@@ -36,4 +36,28 @@ const create = async (link: string, interviewer: string): Promise<s.calendlys.JS
     return db.insert('calendlys', { link, interviewer }).run(pool);
 };
 
-export { create, get };
+/**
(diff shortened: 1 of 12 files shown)


api/db/migrations/003_add_calendly_resource.down.sql
```diff
@@ -0,0 +1,3 @@
+DROP TRIGGER IF EXISTS update_time_calendlys ON calendlys;
+
+DROP TABLE IF EXISTS calendlys;
```
cyrusdiego, line 3: nit: newline

api/src/repositories/calendlyRepository.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (link !== undefined && link !== null) {
+        colOptions.link = link;
+    }
+    if (interviewer !== undefined && interviewer !== null) {
+        colOptions.interviewer = interviewer;
+    }
+    if (interviewee !== undefined && interviewee !== null) {
+        colOptions.interviewee = interviewee;
+    }
+    console.log(db.update('calendlys', colOptions, where).compile());
```
cyrusdiego, line 59: is this console log necessary?

cyrusdiego (approved):
This LGTM! Just left some comments, other than that I think it's good.

armiantos (changes requested):
Blocking because this will change according with changes in #202 as discussed offline

armiantos (approved):
LGTM! Let's get this shipped (after adding authorization to calendly routes hahahahha) 🚀

I'm guessing we'll need to implement some sort of queue and run the matching program on our server as a job instead of having the clients patch the calendly object to avoid race conditions.

calebschoepp:
> LGTM! Let's get this shipped (after adding authorization to calendly routes hahahahha) 
> 
> I'm guessing we'll need to implement some sort of queue and run the matching program on our server as a job instead of having the clients patch the calendly object to avoid race conditions.

Going to merge now. Auth will come in follow up PR.

Matching will be done as a one time thing by us. We can write some sort of sql/ts script to do the job for us.

PR #222, opened 2021-09-06
Add view/download resume

armiantos:
# Overview

* Allow students to view their submitted resume and the reviewed resume

# Changes

* Add non overlaying circular progress bar
* Display non overlaying loading component when loading resume reviews as a student
* Create ResumeReviewViewer which will display their submitted resume (not reviewed)
* Add functionality to download button to download the reviewed resume (as a pdf)

# Questions & Notes

* We don't have an endpoint that can serve the pdf directly from the backend so we have to hack around downloading/viewing pdfs (e.g. appending an `a` component to download the reviewed pdf onClick because we don't want to download all the resume's documents when loading just the resume review list) 
* The viewer still has our App header if we want to remove it, we have to move Header to each of the subpages

# Issue tracking

# Testing & Demo

Test on staging


cyrusdiego (approved):
LGTM!

Issue #231, opened 2021-09-10
Completing a review doesn't automatically refresh the resume list


armiantos:
Solution to #230 should apply to this too

PR #236, opened 2021-09-11
Fix student resume list not reloading

### LINKED ISSUE #230: Submitting a resume for review doesn't automatically reload the resume list

### LINKED ISSUE #232: Race condition on submitting resume review
Implication of #230

armiantos:
# Overview

* Fix resume list not reloading when submitting resume
  * This also stops the race condition that allowed users to submit two resume review requests

# Changes

* Remove redundant fetch in `student/ResumeReview.tsx` (already done in `ResumeList.tsx`)
* Update `shouldReload` when `initiateResumeReview` completes

# Questions & Notes

* I think the root cause of the race condition was the double fetch, will have to check with the fix for #231 

# Issue tracking

Closes #230
Closes #232

# Testing & Demo

When uploading a resume to staging/prod, the resume should directly appear in the list (without a refresh)


cyrusdiego (approved):
Nice and simple fix, good work :)

PR #237, opened 2021-09-11
Fix volunteer resume list reloading too fast

### LINKED ISSUE #231: Completing a review doesn't automatically refresh the resume list

armiantos:
# Overview

* Fix scenario when volunteer completes a resume review but is still listed under their currently reviewing section

# Changes

* Update the thunk to return the `Promise<void>` object
* Group the two fetches together since they're kind of bound to each other via shouldReload

# Questions & Notes

* Root cause was we didn't wait for the asynchronous function to finish so before patch resume review was done, it already triggered the `.fulfilled` event

# Issue tracking

Closes #231

# Testing & Demo

Completing a resume review in staging should now remove the entry in currently reviewing without a refresh

### DIFF
--- a/www/src/routes/volunteer/ResumeReview.tsx
+++ b/www/src/routes/volunteer/ResumeReview.tsx
@@ -21,13 +21,8 @@ const ResumeReview: FC = () => {
     const history = useHistory();
 
     useEffect(() => {
-        if (shouldReload) {
-            dispatch(getAvailableResumeReviews({ tokenAcquirer: getAccessTokenSilently }));
-        }
-    }, [shouldReload]);
-
-    useEffect(() => {
         if (user?.sub !== undefined && shouldReload) {
+            dispatch(getAvailableResumeReviews({ tokenAcquirer: getAccessTokenSilently }));
             dispatch(getReviewingResumeReviews({ tokenAcquirer: getAccessTokenSilently, userId: user.sub }));
         }
     }, [user, shouldReload]);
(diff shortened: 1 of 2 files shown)


cyrusdiego (approved):
LGTM :) just a quick question for my own understanding. Does returning the promise ensure that we wait for the promise to be resolved before proceeding?

armiantos:
> LGTM :) just a quick question for my own understanding. Does returning the promise ensure that we wait for the promise to be resolved before proceeding?

Yes, so `createAsyncThunk` expects an asynchronous function as its second parameter and waits until that asynchronous function to `resolve` before trigger the `.fulfilled` event.

All asynchronous functions return a `Promise` and the way we used it before would imply something like

```
(params, thunkApi) => {
    try {
        patchResumeReview(params);
    } catch (e) {
        return thunkApi.rejectWithValue(e);
    }

    // implicit return
    return new Promise();
}
```
in the above snippet, because `patchResumeReview` is asynchronous, it doesn't have to finish to run the next line which allowed redux to trigger the `fulfilled` event first before the patch actually happens.

```
(params, thunkApi) => {
    try {
        return patchResumeReview(params);
    } catch (e) {
        return thunkApi.rejectWithValue(e);
    }

    // no implicit return
}
```

With the current snippet, redux can wait for the appropriate `Promise` before triggering the `fulfilled` event. The way we do it now is synonymous to 

```
async (params, thunkApi) => {
    try {
        await patchResumeReview(params);
    } catch (e) {
        return thunkApi.rejectWithValue(e);
    }

    // implicit return but if we're here then patchResumeReview has finished because we waited for it
    return new Promise();
}
```

I didn't write it like this because in every other thunk we've written it to return the handler's `Promise`

PR #239, opened 2021-09-12
Close Dialog on switching roles

cyrusdiego:
# Overview

What is the high-level goal of this PR? Is there any important context we should have?

# Changes

- Settings dialog closes when role is changed
- User Profile dialog also closes when role is changed 
    - Not 100% if we want this, I can easily rollback the commit if we want to keep the user profile dialog open when role changes

# Issue tracking

Closes #239 

# Testing & Demo

https://user-images.githubusercontent.com/32345990/132966815-75850cd6-3b30-450f-89af-a92ee558827f.mov

### DIFF
--- a/www/src/components/auth/UserProfile.tsx
+++ b/www/src/components/auth/UserProfile.tsx
@@ -2,15 +2,15 @@ import { useAuth0 } from '@auth0/auth0-react';
 import { Divider, IconButton, Menu, MenuItem } from '@material-ui/core';
 import { fade, makeStyles } from '@material-ui/core/styles';
 import { AccountCircle } from '@material-ui/icons';
-import React, { FC, useRef, useState } from 'react';
+import React, { FC, useRef } from 'react';
 
-import { useAppDispatch } from '../../redux/hooks';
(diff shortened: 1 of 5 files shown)


www/src/components/auth/UserProfile.tsx
```diff
@@ -25,7 +25,7 @@ const UserProfile: FC = () => {
                 color='inherit'
                 ref={userProfileIcon}
                 onClick={() => {
-                    setIsMenuOpen(true);
+                    dispatch(openUserProfileDiaog());
```
armiantos, line 28: Nit: 
```suggestion
                    dispatch(openUserProfileMenu());
```
or
```suggestion
                    dispatch(openUserProfileDialog());
```

armiantos (changes requested):
One nit 🤏

www/src/components/auth/UserProfile.tsx
cyrusdiego: oops missed that typo, ty

armiantos (approved):
LGTM! 🚀

PR #261, opened 2021-09-15
Persist userRole between refreshes

### LINKED ISSUE #255: Volunteer role setting is set back to student role on reload.
Whenever a page reloads, the user role is set back to the student role even though the previous session was as a volunteer.

armiantos:
# Overview

* Allow volunteers to stay as volunteers in the event that they have to refresh a page

# Changes

* Install and integrate redux-persist

# Questions & Notes

* Data is stored in localStorage but the globalStore has no sensitive data other than year and program of the last registered user
* Eliminates the scenario where a student can see another student's resume when they switch back from reviewer to student in the editor

# Issue tracking

Closes #255

# Testing & Demo

As a volunteer, refresh the resume review page, you should stay as a volunteer

### DIFF
--- a/www/package-lock.json
+++ b/www/package-lock.json
@@ -35,6 +35,7 @@
                 "react-router-dom": "^5.2.0",
                 "react-scripts": "4.0.3",
                 "redux": "^4.1.0",
+                "redux-persist": "^6.0.0",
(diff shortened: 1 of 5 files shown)


cyrusdiego (approved):
LGTM!

PR #266, opened 2021-09-19
Prevent double claim of resume review

calebschoepp:
# Overview

Prevent double claim of resume review since UI won't refresh automatically when resumes are claimed.

# Testing & Demo

I want to test this on staging with someone else

### DIFF
--- a/api/src/controllers/resumeReview/patchAllResumeReview.ts
+++ b/api/src/controllers/resumeReview/patchAllResumeReview.ts
@@ -1,6 +1,7 @@
 import { Request, Response } from 'express';
 import type * as s from 'zapatos/schema';
 
+import BadRequestException from '../../exceptions/BadRequestException';
 import InternalServerErrorException from '../../exceptions/InternalServerErrorException';
 import * as resumeReviewRepository from '../../repositories/resumeReviewRepository';
 import controller from '../controllerUtil';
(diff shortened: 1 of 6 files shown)


armiantos:
Tested this locally with 2 simultaneous accounts trying to claim the same resume. Great job!

https://user-images.githubusercontent.com/43416725/133943039-323c31dd-4dca-4b89-9b62-45ac5b61937f.mov

I'll create a PR to put error messages in the front end in the form of [Snackbars](https://mui.com/components/snackbars/)

api/src/controllers/resumeReview/patchAllResumeReview.ts
```diff
@@ -52,6 +53,14 @@ const patchAllResumeReview = controller(async (req: Request<Params, unknown, Req
     await new ParamsValidator().validateAndThrow(req.params);
     await new ReqBodyValidator().validateAndThrow(req.body);
 
+    // Ensure that you can't claim a resume review claimed by someone else
+    if (req.body.reviewer !== undefined) {
+        const rr = await resumeReviewRepository.get(req.params.resumeReview);
+        if (req.body.reviewer !== null && rr[0].reviewer_id !== null) {
+            throw new BadRequestException({ issue: 'You must set the reviewer to null before you can change it to another user.' });
+        }
+    }
```
armiantos, line 62: Nit: I prefer this but this works just fine, because of how we're using the API, `PATCH` with a reviewer field always implies claiming so all good 👍

```suggestion
    if (req.body.state === 'reviewing') {
        const rr = await resumeReviewRepository.get(req.params.resumeReview);
        if (rr[0].reviewer_id !== req.body.reviewer) {
            throw new BadRequestException({ issue: 'You must set the reviewer to null before you can change it to another user.' });
        }
    }
```

armiantos (approved):
Functionality wise LGTM! 🚀

Note for other reviewers, this only affects the patch to claim/unclaim resume reviews because only those requests have a `reviewer` field.

This won't affect the `PATCH` to complete the resume review because that request:
1. Uses the `update_my:resume_reviews` scope instead of `update_all:resume_reviews` which goes through the associated with user check
2. Only updates the `state` of the resume review, it doesn't have the reviewer field

api/src/controllers/resumeReview/patchAllResumeReview.ts
calebschoepp: Good point, since this is functionally the same I'm just going to leave it for now though.

armiantos: Agreed, I'll fix the tests

PR #269, opened 2021-09-19
Fix 500 error for new users

### LINKED ISSUE #268: New users sometimes get 500 on `/me`
Missing return statement sometimes cause a server-side `Cannot set headers after sending to client` error. There's no effect on client side because it still doesn't return content which still redirect new users to the register page. This just improves our logs.

armiantos:
# Overview

* Fix server side 'Cannot set client headers after sending ... to client' error because we don't `return` after `.end()`

# Changes

* Add missing immediate `return` statement in `getMe.ts`

# Questions & Notes

* Other files that call `.end()` already have an implicit `return`

# Issue tracking

Closes #268

# Testing & Demo

* Open dev environment a user that's not registered in your local database


cyrusdiego (approved):
LGTM!

PR #271, opened 2021-09-23
small update to resume page 

cyrusdiego:
# Overview

Give context of resume reviews to students 

image


armiantos (approved):
LGTM! I would put `align='justify'` but that's personal preference (might look a bit too formal with `align=justify` though)

Issue #278, opened 2021-10-23
[Terms of service] Redisplay terms of service dialog


armiantos:
* Prevent users from interacting with the app until they have accepted the terms of service
* Redisplay terms of service dialog when user tries to interact with the app when they haven't agreed to terms of service

PR #286, opened 2021-11-13
Prevent create/update endpoints in backend when user has not agreed to TOS

armiantos:
# Overview

Prevent users from interacting with CompE+'s backend when they have not read the terms of service

# Changes

Created a TOS checker middleware that's applied to all create/update endpoints

# Questions & Notes

This is our last line of defence, I'll be disabling the buttons and displaying the error messages on the front end in another PR. 

Merging this PR does not pose any risk because changes are only done in backend and needs to be promoted to production in heroku before it impacts anyone.

# Issue tracking

Part of #278

# Testing & Demo

Using the new schema try posting a resume for review without having agreed to the terms of service 

image
image

You'll get an infinite loading spinner at the moment but I plan to display an error message where it educates users on where they can click to read/agree the TOS

### DIFF
--- a/api/src/controllers/middleware.test.ts
+++ b/api/src/controllers/middleware.test.ts
@@ -161,7 +161,7 @@ describe('authorizeAndFallthrough middleware', () => {
     });
 });
 
-describe('registeredUser middleware', () => {
+describe('checkTOSAgreement middleware', () => {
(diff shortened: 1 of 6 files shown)


cyrusdiego (approved):
LGTM!

PR #290, opened 2021-11-22
Hide terms of service alert for unauthenticated users

armiantos:
# Overview

Terms of service alert is currently displayed to all users (authenticated or not) because there was no check for this. We only checked if the user has agreed to terms of service and undefined is a falsy value.

Also clear persisted user information when the user is logged out (either through the Logout button or automatically due to JWT expiry)

# Changes

* Add check for authentication status in terms of service alert component
* Simulate clearing the user state when the user is logged out

# Questions & Notes

# Issue tracking

# Testing & Demo

1. From master branch, comment out `dispatch(reset())` in UserProfile
2. Log-in to local.compe.plus
3. Log-out
4. Open resume review page and verify that you have your previous view (you can also double check your local storage and see that the current role is not "")
5. Add the `dispatch(reset())` call to App.tsx when the user is not authenticated
6. Refresh the page and your current role should be cleared

### DIFF
--- a/www/src/App.test.tsx
+++ b/www/src/App.test.tsx
@@ -1,12 +1,13 @@
 import { Auth0ContextInterface, useAuth0, User } from '@auth0/auth0-react';
 import { AsyncThunkAction } from '@reduxjs/toolkit';
 import { render } from '@testing-library/react';
-import { shallow } from 'enzyme';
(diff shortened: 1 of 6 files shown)


cyrusdiego (approved):
LGTM! You beat me to implementing this :') thanks man and good work

